PR #8, opened 2020-04-12
Add tutorial for how to use EyreContext

yaahc:


src/guide/customization.rs
```diff
@@ -0,0 +1,398 @@
...
+//!
+//! First, we have to impl `default` so eyre knows how to construct your context when it creates
+//! your `Report` from a `std::error::Error`:
+//!
+//!
+//! ```rust
+//! fn default(_: &(dyn std::error::Error + 'static)) -> Self {
+//!     Self
+//! }
+//! ```
```
mystor, line 34: Might be worth including the `impl EyreContext` block here, so it's clear where in the code this `default` method should be placed.
I intuited that this was supposed to be a `Default::default`, but it's not.

src/guide/customization.rs
```diff
@@ -0,0 +1,398 @@
...
+//!
+//! # Introduction
+//!
+//! In this tutorial we will go through the process of implementing a custom context step by step:
+//!
+//! * Start with defining an empty context that captures nothing
+//! * Add support for capturing backtraces
+//!     * skip backtrace capture if our source captured one
+//! * Add support for a http status codes
+//!     * Implement a setter for Result<T, Report> with a new Trait
```
mystor, line 11: It'd be nice to have anchor links here for the TOC, to jump to the relevant sections.

src/guide/customization.rs
```diff
@@ -0,0 +1,398 @@
...
+//!
+//!
+//! ```rust
+//! fn default(_: &(dyn std::error::Error + 'static)) -> Self {
+//!     Self
+//! }
+//! ```
+//!
+//! And next we need to implement `debug` to tell eyre how to format the final report including
+//! your custom context:
```
mystor, line 37: I think it'd be good to clarify why this method is called "debug", and how it differs from `fmt::Debug`, perhaps with an intra-doc link to the documentation on the `EyreContext` trait.

src/guide/customization.rs
```diff
@@ -0,0 +1,398 @@
...
+//!
+//!     Ok(())
+//! }
+//! ```
+//!
+//! Things to note, we're using some utilities provided by `eyre` for writing error reports,
+//!
+//! - `Chain`: iterator for errors while the `chain` fn on the error trait is still unstable.
+//! - `Indenter`: a wrapper for formatters that inserts indentation after newlines for handling
+//! multi line error messages nicely
```
mystor, line 66: This is a bit of an awkward place to mention the crates being used. I think it could perhaps be nice to mention them off-hand at the beginning, and have more detailed information in an appendix-type thing?

src/guide/customization.rs
```diff
@@ -0,0 +1,398 @@
+//! A tutorial on how to customize eyre by overriding the Context parameter
+//!
+//! # Introduction
+//!
+//! In this tutorial we will go through the process of implementing a custom context step by step:
```
mystor, line 5: I think it'd be really valuable to have a section before you begin implementing a custom context explaining what a custom context is, and why you might want one. It doesn't necessarily have to be here, but a sentence mentioning where in the docs you can learn the motivation would be nice.

src/guide/customization.rs
```diff
@@ -0,0 +1,398 @@
...
+//!     let mut map = self.map.lock().unwrap();
+//!     let entry = map
+//!         .get_mut(&timeframe)
+//!         .ok_or_else(|| eyre!("No entries for this timestamp"))
+//!         .set_status(StatusCode::NOT_FOUND)?;
+//!
+//!     // ...
+//! }
+//! ```
+//!
```
mystor, line 395: I think including some example code here which reads the status from an error as well would be awesome!

src/guide/customization.rs
```diff
@@ -0,0 +1,398 @@
...
+//! use std::backtrace::BacktraceStatus;
+//!
+//! fn debug(
+//!     &self,
+//!     error: &(dyn std::error::Error + 'static),
+//!     f: &mut core::fmt::Formatter<'_>,
+//! ) -> core::fmt::Result {
+//!     if f.alternate() {
+//!         return core::fmt::Debug::fmt(error, f);
+//!     }
```
mystor, line 156: It'd be good to mention documentation which explains why `alternate` is being checked here.

PR #22, opened 2020-05-23
Fixed use of std::fmt rather than core::fmt

duncanrhamill:
`no_std` seems to be broken because `src/fmt.rs` uses `std` rather than `core` for  `Debug` and `Display` traits. Changed to use `core` which has fixed it for me, all tests still passing.


yaahc:
Ty for the fix! And I'm sorry it took me so long to merge it

PR #24, opened 2020-05-30
[WIP] fix #20 improve test support for `no_std`

### LINKED ISSUE #20: Improve test support for `no_std`
Right now eyre does have support for `no_std` by virtue of the upstream support for no_std errors in `anyhow`. However anyhow never implemented support for testing `no_std`. It would be nice to update the tests to compile and pass when `--no-default-features` is used, however this will require adding `#[cfg(feature = "std")]` all over the test code, and to the doc tests.

mattsse:
This PR is part of the work to fix broken tests in `no_std`. (#20)

So far ea5a0884e5f4a1c8febd5d401ec4fd7c19e2548d includes changes to make eyre compile successfully when `--no-default-feature` is used.
Changes made:
* replaced `std::fmt::Display` with `score::fmt::Display`
* unified the two `Chain` structs into single `Chain` both for `std` and `no_std`. This was necessary because the `crate::chain::Chain` was only `pub(crate)` for `no_std`.
* added some docs and removed unused to make clippy happy.

To fix all the tests for `no_std` is a lot more work, since all the tests using `io::Error` won't work. 
To add test support `no_std`, I could start with replacing the `io::Error` dependent variants in the `TestError` with something that supports `no_std`? This could minimize test duplication for `no_std`
 
Before I proceed I'd like to get some feedback and I'm happy to hear suggestions.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,4 +1,5 @@
 use crate::alloc::Box;
+#[cfg(feature = "std")]
 use crate::chain::Chain;
 use crate::EyreContext;
 use crate::{Report, StdError};
@@ -729,6 +730,7 @@ where
         unsafe { &mut *(self.vtable.object_mut)(self) }
     }
 
+    #[cfg(feature = "std")]
     pub(crate) fn chain(&self) -> Chain<'_> {
         Chain::new(self.error())
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -342,8 +342,8 @@ mod alloc {
     #[cfg(feature = "std")]
     pub(crate) use std::boxed::Box;
 
-    #[cfg(not(feature = "std"))]
-    pub(crate) use alloc::string::String;
+    // #[cfg(not(feature = "std"))]
+    // pub(crate) use alloc::string::String;
 
     // #[cfg(feature = "std")]
     // pub(crate) use std::string::String;
@@ -371,8 +371,11 @@ use core::fmt::Debug;
 #[cfg(feature = "std")]
 use std::error::Error as StdError;
 
+/// `StdError` is a trait representing expectations for error values for no_std mode where the
+/// `std::error::Error` trait is not available.
 #[cfg(not(feature = "std"))]
 pub trait StdError: Debug + Display {
+    /// The lower-level source of this error, if any.
     fn source(&self) -> Option<&(dyn StdError + 'static)> {
         None
     }
@@ -778,8 +781,7 @@ impl EyreContext for DefaultContext {
 ///     None
 /// }
 /// ```
-#[cfg(feature = "std")]
-#[derive(Clone)]
+#[cfg_attr(feature = "std", derive(Clone))]
 #[allow(missing_debug_implementations)]
 pub struct Chain<'a> {
     state: crate::chain::ChainState<'a>,
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -4,13 +4,7 @@ use crate::StdError;
 #[cfg(feature = "std")]
 use std::vec;
 
-#[cfg(feature = "std")]
 pub(crate) use crate::Chain;
-
-#[cfg(not(feature = "std"))]
-pub(crate) struct Chain<'a> {
-    state: ChainState<'a>,
-}
 
 #[derive(Clone)]
 pub(crate) enum ChainState<'a> {


yaahc:
Oh my god when you said `score::fmt::Debug` I was terrified, loool. I've only looked through half of this so far but initial impressions are good, ty for taking this on! And I know the io::Error stuff is a pain in the ass, I tried this once before and eventually just reverted it rather than pushing it to completion.

iirc in no_std eyre doesn't have any way to be constructed from an existing error, it can only be used to construct adhoc error types either via the macros or via `wrap_err`, so I'm not sure if the `TestError` approach will work but if you see something that I missed then feel free to go for it. If that doesn't work out I think its okay to just use `cfg` to disable all of the non-applicable tests in `no_std`.

yaahc:
Also someone else opened a PR to do the std -> core change 6 days ago and I missed it / forgot about it, so I'm gonna merge that PR right after it passes tests, so just a heads up, I know it's a minor inconvenience but I feel like just closing their PR would be a little rude :grimacing:

src/lib.rs
```diff
@@ -342,8 +342,8 @@ mod alloc {
     #[cfg(feature = "std")]
     pub(crate) use std::boxed::Box;
 
-    #[cfg(not(feature = "std"))]
-    pub(crate) use alloc::string::String;
+    // #[cfg(not(feature = "std"))]
```
yaahc, line 345: I think just remove this rather than commenting it out, and the one below if you can.

src/error.rs
```diff
@@ -729,6 +730,7 @@ where
         unsafe { &mut *(self.vtable.object_mut)(self) }
     }
 
+    #[cfg(feature = "std")]
```
yaahc, line 733: I don't think we can make this change because this would mean that the `Item` type of `Chain` changes when we use `no_std` mode. This would cause disabling / enabling features to become a breaking change which is a no-go. Did you need the chain interface on no_std?

mattsse: I added the `cfg` here because the only usage of the `ErrorImpl::chain` function is here: https://github.com/yaahc/eyre/blob/a197feb4e3368f70735a2fe34990fbebb10487cd/src/error.rs#L305-L307 and is marked as `std` only. In addition `ErrorImpl::chain` is `pub crate` making it dead code. It compiles, but clippy rightfully complaints about dead code without this change:

```sh

❯ cargo check --no-default-features                                                                                                                                                                         ⇡!? master 
   Compiling eyre v0.4.2 (/Users/Matthias/git/rust/eyre)
warning: associated function is never used: `chain`
   --> src/error.rs:732:19
    |
732 |     pub(crate) fn chain(&self) -> Chain<'_> {
    |                   ^^^^^
    |
```

mattsse:
:) sorry I scared you with that typo. Unfortunately `score` is already taken, I checked...

I did more research on how to make `cargo test --no-default-features` succeed for the doctests first.

There seems to be two options:
1. the **experimental** [`doc_cfg`](https://doc.rust-lang.org/beta/unstable-book/language-features/doc-cfg.html), which would require adding `!#[feature(doc_cfg)]` to `lib.rs`
2. wrapping the failing doc tests in ` #[cfg_attr(feature = "std", doc = r##...##")]`

I tried 2., there are atm 11 doc tests that fail. Wrapping them in ` #[cfg_attr(feature = "std", doc = r##...##")]` works fine but is kinda weird. So far I fixed only a couple and would like to know what you think about that before I proceed with this cumbersome task -.-.

For https://github.com/yaahc/eyre/blob/a197feb4e3368f70735a2fe34990fbebb10487cd/src/chain.rs#L21-L40
It would look like:

```rust
    /// Construct an iterator over a chain of errors via the `source` method
    ///
    #[cfg_attr(feature = "std", doc = r##"
     # Example

     ```rust
     use std::error::Error;
     use std::fmt::{self, Write};
     use eyre::Chain;
     use indenter::indented;

     fn report(error: &(dyn Error + 'static), f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut errors = Chain::new(error).enumerate();
         for (i, error) in errors {
             writeln!(f)?;
             write!(indented(f).ind(i), "{}", error)?;
         }

         Ok(())
     }
     ```
    "##)]
```

yaahc:
I'd prefer option 1, there are ways around the experimental aspect I think, I've done this same thing in tracing and already opened an issue yesterday to use doc_cfg in eyre.

yaahc:
hey @mattsse, sorry for ignoring this PR for so long, and thank you again for the effort you put into it. I just realized that this PR is no longer something we can merge because I removed support for `no_std` a while ago (and apparently forgot??). I'm really sorry about the confusion.

mattsse:
no worries, I never quite finished it and then also kinda forgot about it  🙈
Thanks for closing!

PR #81, opened 2022-07-07
Fix stacked borrows when dropping

TimDiekmann:
## 🌟 What is the purpose of this PR?

While adding a compatibility layer in `error-stack` for `eyre`, `miri` has detected a bug we also encountered when implementing `error-stack`. 

```
error: Undefined Behavior: trying to reborrow <226876> for Unique permission at alloc95485[0x18], but that tag does not exist in the borrow stack for this location
   --> .cargo/registry/src/github.com-1ecc6299db9ec823/eyre-0.6.8/src/error.rs:521:20
    |
521 |     let unerased = mem::transmute::<Box<ErrorImpl<()>>, Box<ErrorImpl<E>>>(e);
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                    |
    |                    trying to reborrow <226876> for Unique permission at alloc95485[0x18], but that tag does not exist in the borrow stack for this location
    |                    this error occurs as part of a reborrow at alloc95485[0x0..0x30]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
```

## 🔗 Related links

- https://github.com/hashintel/hash/pull/763

## 🔍 What does this change?

- Replacing the `mem::transmute` when dropping with a pointer cast

PR #82, opened 2022-07-07
Run `miri` on CI

TimDiekmann:
As mentioned in #81 here is a small PR disabling the tests, which are not compatible with `miri`. Also, I found a few other `mem::transmute`s. The job will run in the CI, you might want to mark it as required 🙂

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,7 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
           override: true
+          component: miri
       - uses: actions-rs/cargo@v1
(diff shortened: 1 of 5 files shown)


TimDiekmann:
Interestingly, the latest nightly have even more rules for stacked borrows. I tested the branch with `2022-06-23` and it worked. Will investigate it further.

TimDiekmann:
Replacing `Box` with `NonNull` in `error-stack` works. However, I sadly don't have enough time to adjust everything for `eyre` as internally, `eyre` looks quite different. However, if you basically apply the linked PR everything should work. 🙂

PR #92, opened 2023-03-28
unfuck CI (update syn/thiserror)

Gankra:
There's a rocky transition point here around thiserror 1.0.40 and syn2 that depends on precise build-specific dependency-resolution to go awry. Pushing both over the edge in lockstep I think avoids the mess deterministically (although 0.2.11(?) had a bug fix it might be worth forcing too..).


pksunkara:
Superseded by #101. cc @ten3roberts

PR #93, opened 2023-04-20
Add documentation on `wrap_err` vs `wrap_err_with`

akshayknarayan:
This PR adds documentation to address the issue described in https://github.com/yaahc/color-eyre/issues/126.


src/lib.rs
```diff
@@ -1096,6 +1096,13 @@ pub type Result<T, E = Report> = core::result::Result<T, E>;
 ///         # panic!("expected downcast to succeed");
 ///     }
 ///     ```
+///
+/// # `wrap_err` vs `wrap_err_with`
+///
+/// Depending on the instance of [`EyreHandler`] that is installed, using `wrap_err` can have
+/// significant runtime cost (even if the error case does not happen), since [`EyreHandler`] can
+/// incur significant costs (e.g., taking a lock) on error creation. Depending on the context, it
+/// may be significantly more efficient to use `wrap_err_with`.
```
thenorili, line 1105: I find this unclear. I suggest a reword, maybe something like:

```
wrap_err incurs a runtime cost even in the non-error case which wrap_err_with avoids through lazy evaluation. This cost is proportional to the cost of the currently installed EyreHandler's creation step.
```

Can you think of any reason that one would want to eagerly evaluate this error value with wrap_err()? From its description and your analysis I can't imagine a situation where it'd be preferred other than a likely error case or a dispreference for typing out closures.

thenorili (changes requested):

src/lib.rs
akshayknarayan: > I suggest a reword

thanks! your way is clearer. How about the following:

>`wrap_err` incurs a runtime cost even in the non-error case because it requires eagerly constructs the error object. `wrap_err_with` avoids this cost through lazy evaluation. This cost is proportional to the cost of the currently installed [`EyreHandler`]'s creation step.

> Can you think of any reason that one would want to eagerly evaluate this

it seems plausible to have scenarios where the error is already constructed and directly passing it is preferred. for example in a retry loop:
```rust
let mut ok = Err(eyre!("retry loop"));
while ok.is_err() {
  match try_the_thing() {
    Ok(done) => return done,
    Err(latest_err) => ok = ok.wrap_err(latest_err),
  }
}
```

thenorili: Awesome -- for clarity then I'd articulate that for the documentation also so folks know the advantages of both approaches =]

thenorili (approved):
This looks great, thanks so much for this contribution and for being so responsive to feedback :D

thenorili:
This is ready to squash and merge when the merge conflicts are resolved : )

thenorili:
Thanks so much for your contribution again, the new docs look great! :D I'm still getting used to github PRs or I would've merged it myself, I've been using git+reviewboard for years ^^; I hope you have a great weekend.

akshayknarayan:
thanks!

PR #95, opened 2023-05-02
Automatically convert to external errors w/ ensure! and bail!

j-baker:
A pattern documented by thiserror is

```
pub enum MyError {
  ...

  #[error(transparent)]
  Other(#[from] anyhow::Error), // source and Display delegate to anyhow::Error
}
```

It'd be nice for this to work with ensure! but right now, that macro returns an eyre error. There's the 'ScienceError' example but that is more for the case where you want to pick a specific enum variant.

With this PR, the macro additionally runs an .into(). In the case that the return type is an eyre error, obviously .into() will do nothing and be compiled away. In the case that there is a from method, the wrapping will occur. This enables eyre to be used for ergonomic 'implementation detail error' in a thiserror using system which has contractual errors. In the above case, one could write:

```rust
fn foo() -> Result<(), MyError> {
  ensure!(false, "this will be wrapped up in a MyError");
  Ok(())
}
```

To the best of my knowledge, this is a non-breaking change. I also considered modifying `eyre!` to add the .into(), however that seemed more likely to be a breaking change (in particular, breaking `eyre!("foo").into()`!).

### DIFF
--- a/eyre/src/macros.rs
+++ b/eyre/src/macros.rs
@@ -51,13 +51,13 @@
 #[macro_export]
 macro_rules! bail {
     ($msg:literal $(,)?) => {
-        return $crate::private::Err($crate::eyre!($msg));
+        return $crate::private::Err($crate::eyre!($msg).into());
     };
     ($err:expr $(,)?) => {
-        return $crate::private::Err($crate::eyre!($err));
+        return $crate::private::Err($crate::eyre!($err).into());
     };
     ($fmt:expr, $($arg:tt)*) => {
-        return $crate::private::Err($crate::eyre!($fmt, $($arg)*));
+        return $crate::private::Err($crate::eyre!($fmt, $($arg)*).into());
     };
 }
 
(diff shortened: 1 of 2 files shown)


thenorili:
Testing revealed a pattern that this change breaks.

```
     // Tests single-argument `ensure!`
-    let f = || {
+    let f = || -> Result<()> {^M
         ensure!(v + v == 1);
         Ok(())
     };
    assert_eq!(
        f().unwrap_err().to_string(),
        "Condition failed: `v + v == 1`",
    );
```

This test passed before this change, but afterwards it needs a return type annotation.

The compiler warning was surprisingly unhelpful. It suggested something like `Ok::<(), E>` for a generic specified in the function header instead of a return type on the closure.

We should at the very least keep an eye out for any issues opened after this change mentioning a pattern like that so we can tell them how to fix it. I'm not quite sure how to document that behavior though.

thenorili (commented):
Git diff is broken on test_macros.rs because somehow I've added a carriage return to every line. I'll work on doing something about that before I commit. The actual changes are just here.

```
    // Tests single-argument `ensure!`
    let f = || -> Result<()> {
        ensure!(v + v == 1);
        Ok(())
    };
    assert_eq!(
        f().unwrap_err().to_string(),
        "Condition failed: `v + v == 1`",
    );

    // Tests automatically converting to external errors with ensure!()
    let f = || -> Result<(), SomeWrappingErr> {
        ensure!(false, "this will fail");
        Ok(())
    };
    assert!(f().is_err());
}

#[allow(dead_code)]
struct SomeWrappingErr {
    err: eyre::Error,
}

impl From<eyre::Error> for SomeWrappingErr {
    fn from(err: eyre::Error) -> Self {
        SomeWrappingErr { err }
    }
```

thenorili:
Personally I think this is value enough for a very small break. I added a note in the commit message and I'll bring this up with folks before the next release to see how they feel about it, but for now this looks good to merge.

thenorili (approved):

ileixe:
FYI: this PR seems to break API. Now eyre 0.6.10 is not compatible with eyre 0.6.8 as the type is mismatched when user is using this case in closure.

yaahc:
well shit, okay, we'll yank that version for now to give us time to talk about whether or not we want to make that breaking change and pick the right version for the re-release accordingly. Thanks for letting us know.

Issue #98, opened 2023-07-20
Maintenance Status


ten3roberts:
Hello @yaahc . What is the current status of this library?

I've found it to be fantastic to use (especially with `color-eyre`) for constructing rich errors as well as getting a readable panic hook.

However, I've had to go back to anyhow and stop using this library in [flax](https://github.com/ten3roberts/flax) and other libraries due to the Miri/memory invalidations as well as other missing features.

It would be great to have this crate up to speed.

If you do not have the time, which is completely understandable as an open-source maintainer myself, is there any other crate which achieves a similar goal of rich errors and panic messages that you would recommend?

Best regards, Tei

yaahc:
I've been struggling with maintenance on this crate. I still want to keep it up to date and am interested in growing the team to enable others to come in and help maintain it, but I haven't invested enough time into it to make any progress.

Part of the reason I've been having difficulty is that I got really burnt out being in the library team review rotation for a year and a half, and my brain just shuts down whenever I try to do similar work these days. I'm working on recovering from that burnout, but governance work has consistently delayed my efforts to invest more in technical work.

For now, I know that https://docs.rs/error-stack and https://docs.rs/miette both take very similar approaches (miette started as a fork of eyre, not sure how much it has diverged over time but I'm guessing not a ton). I don't want to say that this crate is unmaintained but honestly it might as well be until I get my shit together.

ten3roberts:
Oh that does not sound good, make sure to rest up properly and grow stronger than before.

What sold me on *eyre* was the hook and panic integration, as well as the seamless integration with tracing and backtraces.

I've been absolutely loving my panic messages with `color-eyre` as they are readable, and my code is colored differently from other code so that I can find the cause instantly. I've found the std panics to be extremely hard to read.

Another aspect that I like is that it is simple and uses the standard error trait, whereas miette (a strong contender) uses a custom diagnostics trait which does not necessarily fit your use case, as it is very specific to parsing and source code derived errors and seldom fits other code with the `source` and `url` parts of the trait sticking out like unused but possibly present sore thumbs.

I would love for eyre to get some more love and find maintainers. Is there anything I can do to help in that regard?

pksunkara:
I would also be willing to lend a helping hand.

yaahc:
[...]

PR #99, opened 2023-08-12
Add contributing doc for eyre

yaahc:
Heavily inspired by https://github.com/bevyengine/bevy/blob/main/CONTRIBUTING.md

Fixes https://github.com/eyre-rs/eyre/issues/98

* [Rendered](https://github.com/eyre-rs/eyre/blob/community-hype/CONTRIBUTING.md)

I wanted to keep the docs minimal to start and grow them organically based on need rather than including everything I can think of and making it overwhelming for people, but I'm unsure where to strike the balance. There's a bunch of good stuff from Bevy's contribution guide that I cut that I would love to add over time based on feedback from other contributors.

I think the biggest question right now is how to handle issue tags. Bevy has a comprehensive system similar to the Rust project's issue tags. I want to ensure we have the tags we need to keep things organized, but I'm not sure that we need anything that robust; we don't have nearly as much traffic. I'm leaning towards starting with a minimal set of tags inspired by bevy. The stuff about status (e.g. S-controversial) and difficulty and how those relate to reviews seem like they could be pretty useful for onboarding new reviewers and contributors.

I'll make sure to add all the right tags and check that all the links are working before merging this.

Long term I want to update this or create companion docs to describe various roles like "Triage", "Review", "Merge", etc which may entail rewriting decent chunks of this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,114 +42,60 @@ No matter how you're helping: thank you for contributing to Eyre!
 
 ### Classifying PRs
 
-Our merge strategy relies on the classification of PRs on two axes:
-
-* How controversial are the design decisions
-* How complex is the implementation
+We use labels to organize our issues and PRs.
 
 Each [label](https://github.com/eyre-rs/eyre/labels) has a prefix denoting its category:
 
-* A: Area (e.g. A-Animation, A-ECS, A-Rendering)
+* A: Area or subcrate (e.g. A-eyre, A-color-eyre, A-color-spantrace)
 * C: Category (e.g. C-Breaking-Change, C-Code-Quality, C-Docs)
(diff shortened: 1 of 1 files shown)


pksunkara (approved):
Looks good on first look.

Approving assuming that label and PR examples are moved away from bevy.

ten3roberts (approved):
Looks fantastic.

Really like the clear focus on the fact that anyone is free to contribute and you don't need to be an unsafe Rust zealot to do so.

The part on documentation and constructive criticism is also splendid.

I assume the specific labels will be changed to ones specific to eyre in time.

ten3roberts:
Additionally, just confirming that I am absolutely fine to @mention in PRs. Sometimes you may miss a message due to the volume of mails, so the explicit mention always catches my eye.

PR #101, opened 2023-08-15
Cleanup internals

ten3roberts:
As part of getting to know the codebase I've started to do some spring cleaning

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ thiserror = "1.0"
 trybuild = { version = "1.0.19", features = ["diff"] }
 backtrace = "0.3.46"
 anyhow = "1.0.28"
-syn = { version = "1.0", features = ["full"] }
+syn = { version = "2.0", features = ["full"] }
 pyo3 = { version = "0.13", default-features = false, features = ["auto-initialize"] }
 
 [dependencies]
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -22,7 +22,7 @@ thiserror = "1.0"
 trybuild = { version = "1.0.19", features = ["diff"] }
 backtrace = "0.3.46"
 anyhow = "1.0.28"
-syn = { version = "1.0", features = ["full"] }
+syn = { version = "2.0", features = ["full"] }
```
ten3roberts, line 25: This is what caused the failure of the src/macros.rs

The two versions of syn were conflicting with thiserror

ten3roberts:
Bringing this up to speed with recent changes in rustc using, such as renamed lints.

Feel free to comment

src/macros.rs
```diff
@@ -40,7 +40,6 @@
 ///
 /// # fn main() -> Result<()> {
 /// #     let depth = 0;
-/// #     let err: &'static dyn std::error::Error = &ScienceError::RecursionLimitExceeded;
```
yaahc: I was looking at this for a minute trying to figure out why it was there... honestly still not really sure. I'm guessing it has something to do with asserting that `ScienceError` type implements error, but the `bail!` macro should do the same thing, sooooo.... Seems fine

yaahc (approved):

PR #107, opened 2023-09-27
Move eyre code into a folder

pksunkara:


ten3roberts:
Should the license files be at the top level, and used for most of our crates, or be one per crate?

Most of our crates use the same license, so it'd be useful to merge them together at the top level.

@yaahc what do you think?

pksunkara:
That's what I did. If you see the diff, I haven't changed them but added symlinks to inside the eyre folder.

ten3roberts:
Oh I see, didn't catch that they were symlinks, it is not usually I see them in git

yaahc:
> Should the license files be at the top level, and used for most of our crates, or be one per crate?
> 
> Most of our crates use the same license, so it'd be useful to merge them together at the top level.
> 
> @yaahc what do you think?

I don't see any issue with either version, tho I'm not someone who pays particularly close attention to licensing and legal concerns. Tho I figure if there's an issue someone will open an issue to let us know and we can fix it then.

Issue #89, opened 2022-12-05
Why is `#[no_std]` mentioned in the documentation if there is no `std` feature anymore?


K4rakara:
Title mostly. In PR #29 `#[no_std]` support was removed from the library, but not from the docs.

thenorili:
Thanks for pointing this out =] With your help this issue has been resolved!

thenorili:
Completed in #111

PR #111, opened 2023-11-05
Update documentation on no_std support.

thenorili:
no_std support was removed in 2020 with PR #29. This change updates the documentation to reflect that as suggested in PR #89. It still mentions no_std, which will hopefully keep it on our minds enough to check sometimes and see if the upstream problem has been resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,23 +171,9 @@ avoid using `eyre::Report` as your public error type.
 
 ## No-std support
 
-**NOTE**: tests are currently broken for `no_std` so I cannot guarantee that
-everything works still. I'm waiting for upstream fixes to be merged rather than
-fixing them myself, so bear with me.
-
-In no_std mode, the same API is almost all available and works the same way. To
-depend on Eyre in no_std mode, disable our default enabled "std" feature in
-Cargo.toml. A global allocator is required.
-
-```toml
-[dependencies]
-eyre = { version = "0.6", default-features = false }
-```
-
-Since the `?`-based error conversions would normally rely on the
-`std::error::Error` trait which is only available through std, no_std mode will
-require an explicit `.map_err(Report::msg)` when working with a non-Eyre error
-type inside a function that returns Eyre's error type.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -171,23 +171,9 @@ avoid using `eyre::Report` as your public error type.
...
-eyre = { version = "0.6", default-features = false }
-```
-
-Since the `?`-based error conversions would normally rely on the
-`std::error::Error` trait which is only available through std, no_std mode will
-require an explicit `.map_err(Report::msg)` when working with a non-Eyre error
-type inside a function that returns Eyre's error type.
+No-std support was removed in 2020 in [commit 608a16a] due to unaddressed upstream breakages.
+[commit 608a16a]:
+https://github.com/eyre-rs/eyre/pull/29/commits/608a16aa2c2c27eca6c88001cc94c6973c18f1d5
```
jyn514, line 176: :+1: seems good - if it's not too much work it would be nice to record what needs to be changed upstream to add support back

jyn514 (approved):

README.md
thenorili: I'll ask jane what exactly broke it next time we talk, but I figure it might have shifted over the last three years =]

yaahc: I honestly don't recall, I think it was mostly error in core or the lack there of, and the horrible hacks that we had to work around that.

yaahc (approved):

polarathene:
I came across this from the README and was curious about the problem related to removing the `no_std` support, which seems based on the limited info from above comment to be lack of support for the `Error` trait in `core`?

In [Sep 2024 with Rust 1.81.0](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0/#what-s-in-1-81-0-stable), `core::error::Error` became available, so presumably no horrible hacks are required and `eyre` can bring back such support with the README updated to note a requirement on Rust 1.81+ (_which could also be expressed with the [2024 edition associated to Rust 1.85](https://blog.rust-lang.org/2025/02/20/Rust-1.85.0/), which new projects from this toolchain onwards now default to_).

PR #112, opened 2023-11-05
Move missing_doc_code_examples lint behind its feature gate

thenorili:
CI has been failing due to using the nightly-only
rustdoc_missing_doc_code_examples lint unconditionally.

This moves it behind its feature gate and removes many warnings due to unknown lint.

### DIFF
--- a/color-spantrace/src/lib.rs
+++ b/color-spantrace/src/lib.rs
@@ -60,10 +60,14 @@
 //! [`tracing_error::SpanTrace`]: https://docs.rs/tracing-error/*/tracing_error/struct.SpanTrace.html
 //! [`color-backtrace`]: https://github.com/athre0z/color-backtrace
 #![doc(html_root_url = "https://docs.rs/color-spantrace/0.2.0")]
+#![cfg_attr(
+    nightly_features,
+    feature(rustdoc_missing_doc_code_examples),
+    warn(rustdoc::missing_doc_code_examples)
+)]
 #![warn(
     missing_debug_implementations,
     missing_docs,
-    rustdoc::missing_doc_code_examples,
     rust_2018_idioms,
     unreachable_pub,
     bad_style,
--- a/eyre/src/lib.rs
+++ b/eyre/src/lib.rs
@@ -329,11 +329,14 @@
 //! [`color-spantrace`]: https://github.com/eyre-rs/color-spantrace
 //! [`color-backtrace`]: https://github.com/athre0z/color-backtrace
 #![doc(html_root_url = "https://docs.rs/eyre/0.6.8")]
+#![cfg_attr(
+    nightly_features,
+    feature(rustdoc_missing_doc_code_examples),
+    warn(rustdoc::missing_doc_code_examples)
+)]
 #![warn(
     missing_debug_implementations,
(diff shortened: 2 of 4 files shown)


color-spantrace/src/lib.rs
```diff
@@ -60,10 +60,11 @@
 //! [`tracing_error::SpanTrace`]: https://docs.rs/tracing-error/*/tracing_error/struct.SpanTrace.html
 //! [`color-backtrace`]: https://github.com/athre0z/color-backtrace
 #![doc(html_root_url = "https://docs.rs/color-spantrace/0.2.0")]
+#![cfg_attr(feature = "rustdoc_missing_doc_code_examples",
+            warn(rustdoc::missing_doc_code_examples))]
```
jyn514, line 64: this doesn't actually work. `feature = "..."` is a cargo thing, which is different from "nightly features" (the thing it was erroring about).

afaik the only way to do this like you expect is with hacks like compiling a file in build.rs to see if it supports nightly features, or by manually making people say `cargo doc --features rustdoc_missing_doc_code_examples`.

you should ask jane if she prefers that latter thing, saying people have to use nightly to document (and adding `#![feature()]`, or just removing the warning altogether.

jyn514: some more context - this is what cargo means by "feature" https://doc.rust-lang.org/cargo/reference/features.html
and this is what rustc means https://doc.rust-lang.org/book/appendix-07-nightly-rust.html

jyn514: oh, one last option is to add this warning in CI only with `-Z crate-attr`, but that would be kind of annoying for people building locally since `cargo doc` won't show as many warnings as CI

thenorili:
I've seen examples of feature = "nightly", would that not work instead? I
figured that a specific feature gate would be preferable, thanks for
sharing the distinction between cargo and rustc features :]

jyn514:
`feature = "nightly"` will work if you add it explicitly to Cargo.toml :) people will still need to say `cargo doc --features nightly` though

thenorili:
I've updated the review with an expansion of the existing build script that definitely checks for nightly and a couple small tests to validate that it's working. No new dependencies by the way!

@pksunkara I saw you'd agreed with jyn's critique! I hope you find this update satisfactory :D

pksunkara (approved):

thenorili:
Looks like I didn't run clippy, pardon me! I implemented the suggested change, pushing it now.

eyre/src/lib.rs
```diff
@@ -329,11 +329,17 @@
 //! [`color-spantrace`]: https://github.com/eyre-rs/color-spantrace
 //! [`color-backtrace`]: https://github.com/athre0z/color-backtrace
 #![doc(html_root_url = "https://docs.rs/eyre/0.6.8")]
+// #![cfg_attr(feature = "nightly",
+//             feature(rustdoc_missing_doc_code_examples),
+//             warn(rustdoc::missing_doc_code_examples))]
```
pksunkara, line 334: Why is this still here?

thenorili: ah sorry about that!

PR #113, opened 2023-11-10
Remove obsolete private_in_public lint in nightly.

thenorili:
RFC 2145 is on nightly now, deprecating the public_in_private lint. https://rust-lang.github.io/rfcs/2145-type-privacy.html

public_in_private has been superceded by three new lints. The first two are warn-by-default and the third is allow-by-default. See the excerpt below for some details.

This change skips this lint in nightly in favor of the new warn-by-default lints.

This change revealed a bug in the 'nightly_features' config option for color-spantrace -- it doesn't exist! The lint it was guarding was never turned on. This adds a minimal build script to check for nightly to color-spantrace. Its functionality is tested in the eyre crate, which seems sufficient to me.

The proof this works is that it builds without the warning in nightly : )

After this change there are only two more build warnings for eyre and color-spantrace, the future-incompat dependency warning and the more serious filename collision.

<quote>
    Lint private_interfaces is reported when a type with visibility x is
    used in primary interface of an item with effective visibility y and
    x < y. This lint is warn-by-default.

    Lint private_bounds is reported when a type or trait with visibility
    x is used in secondary interface of an item with effective
    visibility y and x < y. This lint is warn-by-default.

    Lint unnameable_types is reported when effective visibility of a
    type is larger than module in which it can be named, either
    directly, or through reexports, or through trivial type aliases
    (type X = Y;, no generics on both sides). This lint is
    allow-by-default.

    Compatibility lint private_in_public is never reported and removed.
</quote>

### DIFF
--- a/eyre/src/lib.rs
+++ b/eyre/src/lib.rs
@@ -316,10 +316,11 @@
 //! [`color-backtrace`]: https://github.com/athre0z/color-backtrace
 #![doc(html_root_url = "https://docs.rs/eyre/0.6.8")]
 #![cfg_attr(
-    nightly_features,
+    nightly,
     feature(rustdoc_missing_doc_code_examples),
     warn(rustdoc::missing_doc_code_examples)
 )]
+#![cfg_attr(stable, warn(private_in_public))]
 #![warn(
     missing_debug_implementations,
     missing_docs,
@@ -334,7 +335,6 @@
     overflowing_literals,
     path_statements,
     patterns_in_fns_without_body,
-    private_in_public,
     unconditional_recursion,
     unused,
     unused_allocation,
--- a/color-spantrace/build.rs
+++ b/color-spantrace/build.rs
@@ -1,0 +1,53 @@
+use std::env;
+use std::ffi::OsString;
+use std::process::Command;
+use std::str;
+
+fn main() {
+    let version = match rustc_version_info() {
+        Some(version) => version,
+        None => return,
+    };
+    version.toolchain.set_feature();
+}
+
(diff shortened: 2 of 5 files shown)


thenorili:
Missed rust-fmt! I have blocked myself from making the mistake again : )

vscode settings.json
```
    "[rust]": {
        "editor.defaultFormatter": "rust-lang.rust-analyzer",
        "editor.formatOnSave": true,
    },
```

eyre/src/lib.rs
```diff
@@ -320,6 +320,7 @@
     feature(rustdoc_missing_doc_code_examples),
     warn(rustdoc::missing_doc_code_examples)
 )]
+#![cfg_attr(not(nightly_features), warn(private_in_public))]
```
pksunkara, line 323: We need to wait for the next beta as per the warnings you can see below.

thenorili: What warnings are you referring to?

pksunkara: If you check the diff, you can see github comments of clippy warnings in `beta`.

thenorili: Oh! Thank you for pointing that out, I wasn't looking in the right place : ) I'll expand the nightly check to also check for the beta toolchain here.

Do you think it'd be preferable to factor that part of the build script into a library or to keep copying it over?

Also, is there a way you trigger halfway comprehensive CI locally, or do you trigger it on your fork or something?

yaahc (approved):
looks good to me ^_^

Issue #116, opened 2023-11-10
Handle future-incompat dependencies


thenorili:
pyo3 is a direct dependency and can be upgraded to 0.20.0 painlessly.

nom is 3 major versions behind and comes via ansi-parser 0.8.0. I'll reach out to that project and ask for an upgrade =]

`
warning: the following packages contain code that will be rejected by a future version of Rust: nom v4.2.3, pyo3 v0.13.2
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
`

`
- Some affected dependencies have newer versions available.
You may want to consider updating them to a newer version to see if the issue has been fixed.

nom v4.2.3 has the following newer versions available: 5.0.0-alpha1, 5.0.0-alpha2, 5.0.0-beta1, 5.0.0-beta2, 5.0.0-beta3, 5.0.0, 5.0.1, 5.1.0, 5.1.1, 5.1.2, 5.1.3, 6.0.0-alpha1, 6.0.0-alpha2, 6.0.0-alpha3, 6.0.0-beta1, 6.0.0-beta2, 6.0.0-beta3, 6.0.0-beta4, 6.0.0-beta5, 6.0.0, 6.0.1, 6.1.0, 6.1.1, 6.1.2, 6.2.0, 6.2.1, 6.2.2, 7.0.0-alpha1, 7.0.0-alpha2, 7.0.0-alpha3, 7.0.0, 7.1.0, 7.1.1, 7.1.2, 7.1.3
pyo3 v0.13.2 has the following newer versions available: 0.14.0, 0.14.1, 0.14.2, 0.14.3, 0.14.4, 0.14.5, 0.15.0, 0.15.1, 0.15.2, 0.16.0, 0.16.1, 0.16.2, 0.16.3, 0.16.4, 0.16.5, 0.16.6, 0.17.0, 0.17.1, 0.17.2, 0.17.3, 0.18.0, 0.18.1, 0.18.2, 0.18.3, 0.19.0, 0.19.1, 0.19.2, 0.20.0
`

pksunkara:
Because of https://github.com/eyre-rs/color-eyre/issues/80 and https://github.com/eyre-rs/color-eyre/issues/99, we probably would need to move to https://github.com/rust-cli/anstyle and thus not need `ansi-parser` anymore.

sclu1034:
In the meantime, `ansi-parser` has been updated to `nom` v7.

PR #117, opened 2023-11-10
Update future-incompat pyo3 dependency

thenorili:
pyo3 v0.13.2 is tripping future-incompat. This change updates it to the newest 0.20.

The other future-incompat dependency, nom, is an indirect dependency through ansi-parser.

See issue #116

### DIFF
--- a/eyre/Cargo.toml
+++ b/eyre/Cargo.toml
@@ -20,7 +20,7 @@ track-caller = []
 [dependencies]
 indenter = { workspace = true }
 once_cell = { workspace = true }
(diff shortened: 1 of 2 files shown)


thenorili:
I see that the tests with the pyo3 feature are failing here, let me check that locally and send another diff : )

thenorili:
I'll make sure to check relevant features before pushing in the future! looks like pyo3 just renamed pvalue->value in 0.16.0

pksunkara (approved):

Issue #120, opened 2023-11-18
0.6.9 fails to build on unstable-less nightly


Nemo157:
When testing with `RUSTFLAGS=-Zallow-features= cargo +nightly test` to emulate the upcoming stable compiler `eyre` 0.6.9 attempts to use an unstable feature anyway and fails to build

```rust
error[E0725]: the feature `rustdoc_missing_doc_code_examples` is not in the list of allowed features
Error:    --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/eyre-0.6.9/src/lib.rs:320:13
    |
320 |     feature(rustdoc_missing_doc_code_examples),
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Nilstrieb:
This will also break everyone's builds using nighty if the feature ever changes. This is about a lint ans therefore purely internal, using it for everyone seems like a bad idea. Can you just use a feature flag instead that you can enable in your CI?

thenorili:
> This will also break everyone's builds using nighty if the feature ever changes. This is about a lint ans therefore purely internal, using it for everyone seems like a bad idea. Can you just use a feature flag instead that you can enable in your CI?

I don't understand how it would break everyone's build on nightly, could you explain your reasoning please? 


> When testing with RUSTFLAGS=-Zallow-features= cargo +nightly test to emulate the upcoming stable compiler eyre 0.6.9 attempts to use an unstable feature anyway and fails to build

As best I can understand you're saying that it's important to you that we support a nightly toolchain without unstable features, is that correct? I wasn't aware of that use case so I used the nightly toolchain to test for support of unstable features. I'll look into this request in the coming week.

Nemo157:
> As best I can understand you're saying that it's important to you that we support a nightly toolchain without unstable features, is that correct? I wasn't aware of that use case so I used the nightly toolchain to test for support of unstable features.

Yes, I use this in CI on crates that only target the stable compiler as a way to help detect accidental breaking changes in the compiler before they make it to beta. Disabling nightly features is important for that usecase to avoid false positives from changes to unstable features that I don't care about.

thenorili:
https://github.com/rust-lang/rust/blob/master/src/bootstrap/src/bin/rustc.rs
Looks like rustc passes allow-features up for build scripts to inspect. I'll look into adding an opt-in feature flag as well. Do pardon the inconvenience!

[...]

PR #121, opened 2023-11-19
Support unstable-less nightly

thenorili:
This change attempts to put unstable features behind an opt-in feature flag as requested by issue #120. Unstable lints and compile_test.rs are both gated by this feature which is added unconditionally to all CI that touches more than just stable.

As expressed in issue #120, I have some concerns about *needing* to gate compile_test.rs behind feature = "unstable". I'm not certain that this unstable-less nightly eyre will work in production due to the reason that compile_test.rs fails: three of our dependencies, anyhow, thiserror, and proc-macro, all use a test for a nightly toolchain to gate unstable features. Maybe I'm wrong though and this change is sufficient!


thenorili:
oh wow, it looks like public_in_private's been merged on stable!

https://github.com/rust-lang/rfcs/pull/3516

thenorili:
I rebased to my other PR to get rid of the clippy fail, we can remove that before pushing if we want to send this first.

RalfJung:
Cc https://github.com/rust-lang/rust/issues/120804

PR #122, opened 2023-11-19
Remove public_in_private warning

thenorili:
RFC #3516 has been merged and the public in private changes are on stable. It's tripping clippy on stable, so let's fully remove it from our codebase.


thenorili:
Duplicates an ad-hoc addition to #119 : ) Will drop if that merges first.

thenorili:
Superceded by 0.6.9 :D

PR #126, opened 2023-11-25
Add a gitattributes file to specify LF line endings

thenorili:
Since almost all files not on our gitignore are text files, it uses `* text` as a catch-all and specifies *.png as binary.


thenorili:
It would probably make WSL more foolproof to not use auto and just use unix file endings, idk, how would y'all feel about that?

yaahc (approved):

Issue #31, opened 2020-07-10
Converting from anyhow::Error


akshayknarayan:
Hi,
I am porting multiple crates from anyhow to eyre, so I have code that looks something like this:
  
```rust
fn foo() -> Result<(), anyhow::Error> {
  Ok(())
}

fn bar() -> Result<(), eyre::Report> {
  let f = foo().map_err(|e| eyre::Report::new(e))?;
  Ok(())
}
```

This, as expected, gives 
```
error[E0277]: the trait bound `anyhow::Error: std::error::Error` is not satisfied
  --> src/lib.rs:6:49
   |
6  |     let f = foo().map_err(|e| eyre::Report::new(e))?;
   |                                                 ^ the trait `std::error::Error` is not implemented for `anyhow::Error`
   |
  ::: /home/akshayn/.cargo/registry/src/github.com-1ecc6299db9ec823/eyre-0.6.0/src/error.rs:21:12
   |
21 |         E: StdError + Send + Sync + 'static,
   |            -------- required by this bound in `eyre::error::<impl eyre::Report>::new`

error: aborting due to previous error
```

Since `anyhow::Error` and `eyre::Report` are so similar, is there a better way to construct an `eyre::Report` mid-stack than this, which is sad and throws out the context:
```rust
fn foo() -> Result<(), anyhow::Error> {
  Ok(())
}

fn bar() -> Result<(), eyre::Report> {
  let f = foo().map_err(|e| eyre::eyre!("{:?}", e))?;
  Ok(())
}
```

Thanks for the help!

yaahc:
the best way to convert between the two would be to first convert the anyhow::Error to a Box<dyn Error> and then convert that to an Eyre report with the macro. that way you'll preserve the chain of sources at least, though you'll still loss the backtrace that anyhow captured originally.

akshayknarayan:
that makes sense, thanks!

ppoliani:
> ```rust
> map_err(|err| eyre!(Box::new(err))
> ```

Alternatively you can wrap this inside a declarative macro

```rust
map_err!(foo());

#[macro_export]
macro_rules! map_err {
  ($fun:expr) => {
    $fun.map_err(|e| eyre!(Box::new(e)))?
  }
}
```

pksunkara:
@yaahc Please let me know if you welcome a PR

yaahc:
> @yaahc Please let me know if you welcome a PR

[...]

Issue #65, opened 2021-12-06
Document how to wrap an `eyre::Report` with a different error type


TheButlah:
Wrapping a source error type as the cause of an `eyre::Report` is very straightforward with the `wrap_err` function. But suppose I have an `eyre::Result<()>` and want to convert it into a `Result<(), ()>` without discarding the helpful backtrace/spantrace from `eyre`? Is that even possible?

If it is, can it be documented?

yaahc:
> But suppose I have an `eyre::Result<()>` and want to convert it into a `Result<(), ()>` without discarding the helpful backtrace/spantrace from `eyre`? Is that even possible?

As stated, no, you definitely cannot convert an `eyre::Result<()>` to a `Result<(), ()>` without losing all relevant information. I think I may be misunderstanding what you mean, but based on what you wrote, you're essentially asking for `Result<(), eyre::Report> -> Result<(), ()>` which can be reduced to `eyre::Report -> ()`, aka converting `eyre::Report` to the [unit type](https://doc.rust-lang.org/stable/std/primitive.unit.html). This a `drop`, and can only be accomplished by discarding the original `eyre::Report`. All relevant information to an error such as backtrace and spantrace are stored in the `eyre::Report`, not the `Result`. We cannot discard this `Report` type without losing all of that helpful information.

TheButlah:
> you're essentially asking for Result<(), eyre::Report> -> Result<(), ()> which can be reduced to eyre::Report -> (), aka converting eyre::Report to the unit type

Yeah that sums it up pretty well and it makes sense why that would not be viable. I guess what I'm really looking for is a way to continue to have whatever error type `E` my Result is supposed to use, but still propagate the helpful backtrace info that eyre contains. I can create a new `Report` but since its not a `Report<E>` I cannot access the wrapped error easily.

yaahc:
I'm not sure I understand. Can you give me an example of what you're trying to do?

TheButlah:
Wow I left this for a full year lol.

I believe what I was asking for was a way to propagate the information about the chain of underlying errors to other error error libs like thiserror.

I wanted to be able to use `eyre` in a library for prototyping easy errors, and then have a parent library that uses thiserror that handles error from all subsystems, some of which may be using `eyre`. But I didn't want to lose the helpful error chain.

PR #127, opened 2023-11-29
Structural conversion from anyhow errors

ten3roberts:
This implements a structural conversion to convert betwen other errors kinds and eyre *Reports*.

This allows richer error types to provide more context to the eyre report, such as anyhow by preserving backtrace and other context where possible.

This does not apply for anyhow, but future crates as well by implementing `IntoEyreReport`. This can also be used to provide a conversion from a custom error type or enum which contains a `Report` that does not nest the Report within a new report, but flattens it which ensures the spantrace and backtrace are forwarded instead of being debug printed without structure inside another backtrace, which can be very confusing.

This is an issue I've had with [flax](https://github.com/ten3roberts/flax), where the library error enum has a variant for a boxed user error, as well as concrete error variants. I wanted to avoid nesting a `eyre::Report` or `Box<dyn Error>` inside another Report if a user wraps flax's error themselves.

https://github.com/eyre-rs/eyre/issues/65#issuecomment-1248085692

Resolves: #31

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,3 +15,5 @@ rust-version = "1.65.0"
 [workspace.dependencies]
 indenter = "0.3.0"
 once_cell = "1.18.0"
+# For use with anyhow-compat
+anyhow = "1.0"
--- a/eyre/Cargo.toml
+++ b/eyre/Cargo.toml
@@ -16,11 +16,14 @@ rust-version = { workspace = true }
 default = ["auto-install", "track-caller"]
 auto-install = []
 track-caller = []
+anyhow-compat = ["anyhow"]
+anyhow-compat-backtrace = ["anyhow-compat", "anyhow/backtrace"]
 
 [dependencies]
 indenter = { workspace = true }
 once_cell = { workspace = true }
 pyo3 = { version = "0.20", optional = true, default-features = false }
+anyhow = { workspace = true, optional = true }
 
 [dev-dependencies]
 futures = { version = "0.3", default-features = false }
--- a/eyre/src/builder.rs
+++ b/eyre/src/builder.rs
@@ -1,0 +1,79 @@
+use crate::{
+    backtrace::Backtrace,
+    vtable::{
+        object_boxed, object_downcast, object_downcast_mut, object_drop, object_drop_front,
+        object_mut, object_ref, ErrorVTable,
+    },
+    Report, StdError,
+};
+use std::fmt::Display;
+
+#[derive(Default)]
+pub struct ReportBuilder {
+    backtrace: Option<Backtrace>,
+}
+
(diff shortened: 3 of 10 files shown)


ten3roberts:
Screenshot 2023-11-29 at 12 53 03

ten3roberts:
> the main rough edge to watch out for is accidentally using `Into` instead of `IntoReport` and losing context.
> 
> you might be able to catch these at runtime with debug asserts inside of the eyre hook when converting to a report, by inspecting the `&dyn Error` to look for `eyre::Reports` or other known types you want to use `IntoReport` on.

Yes, this is a footgun and unfortunately as far as I can tell a limitation with Rust.

For reference we discussed this a bit yesterday, and while the `IntoEyre` trait will allow you to convert types to `eyre::Report`s in the absence of `std::error::Error`, such as `#![no_std]`, it will also allow you to use `?` when the Error type *does* implement error, in which case it will skip the specialized flattening and backtrace re-attaching for e.g errors containing eyre reports themselves, such as https://github.com/ten3roberts/flax/blob/main/src/error.rs.

Something we *could* do, is try and rethink the `impl<T> FromT> for eyre::Report where T: Error` and see if we can work around that one, as it is the root cause of the conflicting traits that disallow implementing `Into<eyre::Report>`.

Any ideas are highly welcome ❤️

thenorili:
> Some further consideration and ideas regarding this:
> 
> 1. We allow `std::error::Error` and `IntoEyre` to coexist, and while the user can use `?` and lose some context, they can go back to the code and change it into `into_eyre` to get more information when needed, similar to how one today goes back and adds or adds a `.wrap_err`, or gives it more information.
> 2. We modify the `eyre!` to firstly look for an `IntoEyre` implementation, before falling back on `Error` or `Display`
> 3. We create a conflicting impl with Error to force downstream users to implement one or the other, not both. This is most disruptive and may make interop harder if your type can't _also_ be a normal error. This would be detrimental for flax.
> 4. Modify downstream eyre handlers to check for nested eyre Reports. This is very unstable and will be prone for false-positives or impossibilties where the type system is not capable of using `IntoEyre`, most commonly because an error arrives as a `Box<dyn Error>` or `impl Error`, but the current concrete type happens to contain an eyre Report or implement `IntoEyre`. This would be a nightmare in stability and leakage of error internals for erased types, and it would also not catch all since an `IntoEyre` may do more than just flattening, and we could not detect that and we'd miss missing uses of `IntoEyre`.

I'm of the same mind as Jane, 1+2 sounds best. I think there's a compromise with 3. We could, in theory, publish a non-default lint that checks for usage of ? or into() where into_eyre() is also possible. That way folks who do care about not losing context have that option while the general case stays flexible and nondisruptive.

Issue #125, opened 2023-11-21
Extend `Option` with `ok_or_eyre`?


LeoniePhiline:
When discussing https://docs.rs/eyre/latest/eyre/index.html#context-and-option, a user remarked that a closure and macro invocation just for a static string error (as documented: `opt.ok_or_else(|| eyre!("new error message"))`) seemed cumbersome.

The reasoning at the documentation's "`Context` and `Option`" section does make sense - `WrapErr` would indicate wrapping an error, where there is (literally) `None`. This issue is _not_ asking to implement `WrapErr` for `Option`.

An extension trait, offering `ok_or_eyre` on the `Option` type does seem useful to avoid closure + macro boilerplate:

```rust
use std::fmt::{Debug, Display};

pub trait OptionExt<T> {
    /// Example usage:
    ///
    /// ```rust
    /// let option: Option<()> = None;
    ///
    /// let result = option.ok_or_eyre("static str error");
    ///
    /// assert_eq!(result.unwrap_err().to_string(), "static str error");
    /// ```
    fn ok_or_eyre<M>(self, message: M) -> crate::Result<T>
    where
        M: Debug + Display + Send + Sync + 'static;
}

impl<T> OptionExt<T> for Option<T> {
    fn ok_or_eyre<M>(self, message: M) -> crate::Result<T>
    where
        M: Debug + Display + Send + Sync + 'static,
    {
        match self {
            Some(ok) => Ok(ok),
            None => Err(crate::Report::msg(message)),
        }
    }
}
```

Is this something you would consider accepting as contribution? Names are of course up for bikeshedding.

yaahc:
yes! I think this is a fantastic way to navigate that concern; I'd definitely not object to this addition.

LeoniePhiline:
Landed with #129's merge.

PR #129, opened 2023-12-06
Extend `Option` with `ok_or_eyre`

LeoniePhiline:
Previously, a closure and macro invocation was
required to generate a static string error object
from an `Option::None`.

This change adds an extension trait, providing
the `ok_or_eyre` method on the `Option` type.

`Option::ok_or_eyre` accepts static error messages
and creates `Report` objects lazily in the `None` case.

Implements #125

### DIFF
--- a/eyre/tests/test_option.rs
+++ b/eyre/tests/test_option.rs
@@ -1,0 +1,15 @@
+mod common;
+
+use self::common::maybe_install_handler;
+use eyre::OptionExt;
+
+#[test]
+fn test_option_ok_or_eyre() {
+    maybe_install_handler().unwrap();
+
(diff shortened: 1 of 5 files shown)


eyre/tests/test_option.rs
```diff
@@ -0,0 +1,15 @@
+mod common;
+
+use self::common::maybe_install_handler;
+use eyre::OptionExt;
+
+#[test]
+fn test_option_or_eyre() {
```
LeoniePhiline, line 7:
```suggestion
fn test_option_ok_or_eyre() {
```

Should fixing this name be worth another workflow run?

PR #138, opened 2023-12-20
feat: introduce an `"anyhow"` compatibility layer feature flag

### LINKED ISSUE #131: Ergonomics: Hide `context` behind a feature
When using `eyre` day by day, rust analyzer often proposes to import the `Context` trait, when `WrapErr` functions like `wrap_err_with` are not imported.

(Rust analyzer does not even offer importing `WrapErr`, but only `Context`. I do not know why.)

As I do not need `anyhow` compatibility, I would prefer `mod context` to be feature gated.

LeoniePhiline:
This change hides the `anyhow` compatibility layer behind an `"anyhow"` feature flag.
In `eyre` v0.6 the feature is enabled by default.

Fixes #131

### DIFF
--- a/eyre/src/context.rs
+++ b/eyre/src/context.rs
@@ -1,5 +1,5 @@
 use crate::error::{ContextError, ErrorImpl};
-use crate::{ContextCompat, Report, StdError, WrapErr};
+use crate::{Report, StdError, WrapErr};
 use core::fmt::{self, Debug, Display, Write};
 
 #[cfg(backtrace)]
@@ -62,6 +62,7 @@ where
         }
     }
 
+    #[cfg(feature = "anyhow")]
     fn context<D>(self, msg: D) -> Result<T, Report>
     where
         D: Display + Send + Sync + 'static,
@@ -69,6 +70,7 @@ where
         self.wrap_err(msg)
     }
 
+    #[cfg(feature = "anyhow")]
     fn with_context<D, F>(self, msg: F) -> Result<T, Report>
     where
         D: Display + Send + Sync + 'static,
@@ -78,7 +80,8 @@ where
     }
 }
 
-impl<T> ContextCompat<T> for Option<T> {
+#[cfg(feature = "anyhow")]
+impl<T> crate::ContextCompat<T> for Option<T> {
     fn wrap_err<D>(self, msg: D) -> Result<T, Report>
     where
         D: Display + Send + Sync + 'static,
--- a/eyre/src/error.rs
+++ b/eyre/src/error.rs
@@ -116,6 +116,7 @@ impl Report {
         unsafe { Report::construct(error, vtable, handler) }
     }
 
+    #[cfg(feature = "anyhow")]
     #[cfg_attr(track_caller, track_caller)]
     pub(crate) fn from_display<M>(message: M) -> Self
     where
--- a/eyre/src/lib.rs
+++ b/eyre/src/lib.rs
@@ -265,7 +265,17 @@
 //! vice-versa by re-exporting all of the renamed APIs with the names used in
 //! `anyhow`, though there are some differences still.
 //!
-//! #### `Context` and `Option`
+//! ### Disabling the compatibility layer
+//!
+//! The `anyhow` compatibility layer is enabled by default.
+//! If you do not need anyhow compatibility, it is advisable
+//! to disable the `"anyhow"` feature:
+//!
+//! ```toml
+//! eyre = { version = "0.6", default-features = false, features = ["auto-install", "track-caller"] }
+//! ```
+//!
+//! ### `Context` and `Option`
 //!
 //! As part of renaming `Context` to `WrapErr` we also intentionally do not
 //! implement `WrapErr` for `Option`. This decision was made because `wrap_err`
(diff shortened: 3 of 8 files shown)


eyre/src/context.rs
```diff
@@ -78,6 +82,7 @@ where
     }
 }
 
+#[cfg(feature = "anyhow")]
 impl<T> ContextCompat<T> for Option<T> {
```
ten3roberts, line 86: Maybe this could use `crate::ContextCompat` directly to reduce the number of `cfg`, as they are oftentimes harder to read in imports

eyre/src/error.rs
```diff
@@ -116,6 +116,7 @@ impl Report {
         unsafe { Report::construct(error, vtable, handler) }
     }
 
+    #[cfg(feature = "anyhow")]
```
ten3roberts, line 119: Is the removal of `from_display` intended, as it is useful of conversion of non-error errors such as string or other display structs.

I've run into this for `no-std` crates, which do not implement `std::error::Error` for their exported error type, but rather just `Display + Debug`

eyre/src/lib.rs
```diff
@@ -265,7 +265,17 @@
 //! vice-versa by re-exporting all of the renamed APIs with the names used in
 //! `anyhow`, though there are some differences still.
 //!
-//! #### `Context` and `Option`
+//! ### Disabling the compatibility layer
```
ten3roberts, line 268: Great.

This further goes into the general guideline that library authors should default to `default_features = false`

eyre/tests/test_location.rs
```diff
@@ -127,6 +130,7 @@ fn test_option_compat_wrap_err() {
     println!("{:?}", err);
 }
 
+#[cfg(feature = "anyhow")]
```
ten3roberts, line 133: Could we change these to use `eyre::WrapErr` or `ok_or_eyre` instead to test them, and show there is a non-anyhow alternative?

ten3roberts (changes requested):

eyre/src/error.rs
LeoniePhiline: `from_display` is `pub(crate)` and only used in the `anyhow` compatibility layer.

I would think changing it to `pub` is outside the scope of this PR.

ten3roberts: Ok got it, thought it was also used by the macro, but it used the `from_adhoc` instead

eyre/tests/test_location.rs
LeoniePhiline: I am not sure I understand what you mean here.

To my mind, this test tests the interaction of `wrap_err_with` with `ContextCompat`.

eyre/src/context.rs
LeoniePhiline: Fixed in https://github.com/eyre-rs/eyre/commit/3237c3895ebbc29ef83e47a7b49cbfb34205b2ee

eyre/tests/test_location.rs
LeoniePhiline: On a second look, this specifically tests `Option::wrap_err_with`, which is not provided by `WrapErr`.

`WrapErr` on `Result` is tested in [`fn test_wrap_err`](https://github.com/eyre-rs/eyre/pull/138/files#diff-1ff47dac6cf55e34ff587968c5b1f1ec6b6ae39d2668a66ecba3633163a21fc5R43).

`OptionExt` provides `Option::ok_or_eyre`, which is the pure-`eyre` analog to `Option::wrap_err_with` from the `anyhow` compat layer.

ten3roberts: Yes, I thought it may be worthwhile to show how to do the same without anyhow in another test.

Just realized that the `ContextCompat` has a `.context` method and a `.wrap_err`, which got me into thinking it just called down to `WrapErr::wrap_err` instead; which is rather inconsistent with two traits with the same method name. All the better to get this feature gated.

I can push up a test later that adds a test for `ok_or_eyre` for location, as I realized that it does not pass currently 😬 

I am not 100% familiar with all of this crate yet, so thank you for your patience

LeoniePhiline: Ah, now I see! :) 

Testing `ok_or_eyre` for location should have been part of https://github.com/eyre-rs/eyre/pull/129. I totally missed that. I, too, am not 100% familiar with all of this crate yet, and thank *you* for your patience just the same.

ten3roberts: https://github.com/eyre-rs/eyre/pull/140

Fixed it here, found a missing `track_caller`

LeoniePhiline: Awesome, thank you!

ten3roberts (approved):
Awesome!

@yaahc is there any function, trait, etc here that you see which shouldn't be behind this flag, or does it all look good?

Nemo157:
Note, this is a breaking change, anyone using `default-features = false` and the combat layer will be broken.

LeoniePhiline:
@Nemo157 Yes, this is planned for a breaking release.

ten3roberts (changes requested):
There is only one way around this breaking change.

Making the feature additive but negative, I.e `no-anyhow-compat` would work, but will break dependencies with the feature resolver.

As such, the only viable option would be to keep this as a positive feature flag, and improve it in the next release. Potentially, we can default this to off but this is yet to be decided.

To merge this now we could create a breaking changes branch and merge it into there, and then backport bugfixes to the current 0.6 release on `main`

ten3roberts:
Certainly, a maintenance would work and we can backport all bug fixes from main to 0.6 to prevent them being forgotten.

That is correct, there is nothing that would need to be done on your part and the feature flag is really good 😊. I spoke to Jane and we didn't find that there was any item that we missed to include under the feature flag.

ten3roberts (approved):
Awesome.

These changes will now land on master, which prepares for the next breaking release.

As Nemo pointed out, this is because disabling behavior by moving it to a default feature is a breaking change.

ten3roberts:
Absolutely, that would be great.

It would further go into our intents of making the error wrapping represent a logical error chain, rather than context attached to an error. The anyhow compatibility has unfortunately incentivized the latter approach where through API naming the error chains are treated like context stores where you can interpolate arbitrary data, such as usernames or other variables.

Making it opt-in would thus be awesome, and push towards a more descriptive chain of errors.

Do you mind opening a PR with a short explanation in the Readme of what it means and how to opt in/how it compares to anyhow?

LeoniePhiline:
Yes, I can do that!

PR #140, opened 2023-12-20
fix: `ok_or_eyre` not using `track_caller`

ten3roberts:
`ok_or_eyre` did not use `track_caller`, which meant the error location was incorrect compared to the equivalent anyhow::context feature


LeoniePhiline:
Thanks, good catch!

thenorili (approved):
Good catch, I'll keep a close eye out for this attribute on future reviews.

Issue #141, opened 2023-12-20
Segfault in `Report::downcast()` introduced in 0.6.9


stevepryde:
I recently had a test segfault after updating to the latest eyre (0.6.11). The same test passes on 0.6.8 but segfaults on 0.6.9 or later.

```
error: test failed, to rerun pass `--bin ...`

Caused by:
  process didn't exit successfully: ... (signal: 11, SIGSEGV: invalid memory reference)
```

Here is a minimal reproduction:
```rust
use eyre::Context;

fn main() {
    let err = trigger().unwrap_err();
    let _ = err.downcast::<MyError>();
}

#[derive(Debug, thiserror::Error)]
enum MyError {
    #[error("error building http client")]
    ClientConstruction,
    #[error(transparent)]
    SsmError(#[from] aws_sdk_ssm::Error),
}

fn trigger() -> eyre::Result<()> {
    do_thing().wrap_err("failed")?;
    Ok(())
}

fn do_thing() -> Result<(), MyError> {
    Err(MyError::ClientConstruction)
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_me() {
        let err = trigger().unwrap_err();
        let _ = err.downcast::<MyError>();
    }
}
```

Cargo.toml deps
```toml
[dependencies]
eyre = "0.6.9" # or later
thiserror = "1"
aws-sdk-ssm = "1"
```

I included the test because `cargo test` gives slightly more info.

Something about the second error variant is related, even though that variant isn't constructed.
I used `aws_sdk_ssm::Error` in the error variant because that's how I hit the issue.

The issue seems to be in src/error.rs, in `context_drop_rest()`, which was changed in 0.6.9. Any reason why this code differs from `anyhow` ?

stevepryde:
Additional details:
* Mac M1 (macOS 14.2.1)
* Rust 1.74.1 (also tried on 1.74.0 and 1.73.0, same result)

stevepryde:
I get the same segfault also on latest Manjaro Linux (x86-64), using Rust 1.74.1.

abusch:
I've managed to simplify the reproduction case to remove dependencies (`thiserror` and `aws-sdk-ssm` are off the hook):

```rust
#![allow(dead_code)]
use std::{error::Error, fmt::Display};

use eyre::Context;

fn main() {
    let err = trigger().unwrap_err();
    let _ = err.downcast::<MyError>();
    println!("it worked!");
}

[...]

PR #143, opened 2023-12-29
Fix invalid drop impl call in `Report::downcast`

ten3roberts:
The context chain downcast called converted to the wrong inner type which caused the wrong drop impl to be called.

The tests did not catch this because they had compatible drop implementations due to matching type layout. Solved by swizzling the fields of the chain test types to force incompatible layouts

Resolves: #141

### DIFF
--- a/eyre/src/error.rs
+++ b/eyre/src/error.rs
@@ -698,13 +698,13 @@ where
     // ptr::read to take ownership of that value.
     if TypeId::of::<D>() == target {
         unsafe {
-            e.cast::<ErrorImpl<ContextError<ManuallyDrop<E>, E>>>()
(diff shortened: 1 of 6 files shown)


dtolnay (approved):
This seems worth publishing an advisory for: https://github.com/rustsec/advisory-db

thenorili:
Great catch, and clever work on the test!

thenorili (approved):
Looks good to me, excellent work debugging the test.

stevepryde:
Any reason why this fix hasn't been released yet (and affected versions yanked)? It's a known soundness issue in a popular crate.

thenorili:
Thanks for pointing that out. There were some cross wires and folks thought it had made it into the last release when it had not. Working on it!

PR #145, opened 2024-01-04
Rename eyre::Ok to eyre::EyreOk to solve conflicts

### LINKED ISSUE #144: `fn Ok()` breaks all code with `use eyre::*` in new releases
I noticed a breaking change with `Ok()`, meaning that all code like
```rust
match result {
  Ok(ok) => ...,
  Err(err) => ...
}
```
will break.
```log
error[E0532]: expected tuple struct or tuple variant, found function `Ok`
   |
68 |                 Ok(rows) => rows,
   |                 ^^ not a tuple struct or tuple variant
   |
help: consider importing one of these items instead
   |
1  + use core::result::Result::Ok;
   |
1  + use pin_utils::core_reexport::result::Result::Ok;
   |
1  + use serde::__private::Ok;
   |
1  + use std::result::Result::Ok;
   |
     and 1 other candidate

```

This is not acceptable as Ok being a very common prelude item in Rust.

Even worse, the following code in the same file can't be ergonomic with `fn Ok`
```rust
let closure = || /* do something */ Ok(());

match result {
  Ok(ok) => ...,
  Err(err) => ...
}
```

JakkuSakura:
closes #144

### DIFF
--- a/eyre/src/lib.rs
+++ b/eyre/src/lib.rs
@@ -1272,7 +1272,7 @@ pub trait ContextCompat<T>: context::private::Sealed {
 ///    |         consider giving this pattern the explicit type `std::result::Result<i32, E>`, where the type parameter `E` is specified
 /// ```
 #[allow(non_snake_case)]
-pub fn Ok<T>(t: T) -> Result<T> {
+pub fn EyreOk<T>(t: T) -> Result<T> {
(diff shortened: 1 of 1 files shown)


thenorili:
Closing per the discussion in #144

Issue #147, opened 2024-01-24
`OptionExt` feature test is incorrectly feature-gated behind the `anyhow` flag


LeoniePhiline:
Introduced at https://github.com/eyre-rs/eyre/commit/34bd1d98937d775dc8916a638d7ac0eece9e6e97#diff-1ff47dac6cf55e34ff587968c5b1f1ec6b6ae39d2668a66ecba3633163a21fc5R86

Issue #136, opened 2023-12-20
[breaking] Remove `anyhow` from default features


LeoniePhiline:
#131 introduces a new `anyhow` default feature flag to allow users to hide aliased exports such as `eyre::Error` and `eyre::Context`.

This issue tracks the removal of `anyhow` from the list of default features, making the compatibility layer with the `anyhow` crate opt-in.

LeoniePhiline:
**Update:** This change is planned to go into the next breaking release, since [introducing a (default) `anyhow` feature flag](https://github.com/eyre-rs/eyre/issues/131) ended up being an [unintended breaking change](https://github.com/eyre-rs/eyre/pull/138#issuecomment-1870757313).

@ten3roberts [responded](https://github.com/eyre-rs/eyre/pull/138#issuecomment-1899256943) to my question _"Since it's a breaking change anyway, [shall we then remove it from default features right away](https://github.com/eyre-rs/eyre/pull/138#issuecomment-1894161838)?"_ with:

> Absolutely, that would be great.
> 
> It would further go into our intents of making the error wrapping represent a logical error chain, rather than context attached to an error. The anyhow compatibility has unfortunately incentivized the latter approach where through API naming the error chains are treated like context stores where you can interpolate arbitrary data, such as usernames or other variables.
> 
> Making it opt-in would thus be awesome, and push towards a more descriptive chain of errors.
> 
> Do you mind opening a PR with a short explanation in the Readme of what it means and how to opt in/how it compares to anyhow?

LeoniePhiline:
Status: Waiting for https://github.com/eyre-rs/eyre/issues/149 due to some thematic overlap.

LeoniePhiline:
Status: The previous blocker #149 has been merged. #148 is ready for review and merge.

LeoniePhiline:
Status: The previous blocker #149 has been merged. #180 is ready for review and merge.

PR #148, opened 2024-01-24
fix: remove `anyhow` feature flag from `OptionExt` location test 

LeoniePhiline:
# 1. Clarify trait usage in location test

[`8522f02`](https://github.com/eyre-rs/eyre/pull/148/commits/8522f02518330d4a0178a0e940c4a910bb9299e0) 

Both `WrapErr` and `ContextCompat` expose `anyhow` compatibility methods.

- `ContextCompat` is completely feature gated behind the `anyhow` flag.

- `WrapErr`, on the other hand, feature-gates individual functions behind the `anyhow` flag.

This has led to [confusion][confusion] in the past.

This change moves the `use eyre::WrapErr` statement from the top of the module into each individual test, as [identically named functions](https://github.com/eyre-rs/eyre/issues/149) (`wrap_err.*`) are exposed both by `WrapErr` and `ContextCompat`.

With `use eyre::WrapErr` moved directly into the tests, it becomes clear which trait-provided `fn` is being called.

[confusion]: https://github.com/eyre-rs/eyre/pull/138#discussion_r1432909562

# 2. Remove `anyhow` feature flag from `OptionExt` location test 

[`68744f1`](https://github.com/eyre-rs/eyre/commit/68744f19fb5f755aafdee2b163510b2e4e1fca37) 

This bug was introduced (by me) in https://github.com/eyre-rs/eyre/commit/34bd1d98937d775dc8916a638d7ac0eece9e6e97#diff-1ff47dac6cf55e34ff587968c5b1f1ec6b6ae39d2668a66ecba3633163a21fc5R86 - partly due to the confusion fixed with the above commit.

Fixes https://github.com/eyre-rs/eyre/issues/147

### DIFF
--- a/eyre/tests/test_location.rs
+++ b/eyre/tests/test_location.rs
@@ -83,7 +83,6 @@ fn test_wrap_err_with() {
     println!("{:?}", err);
 }
 
-#[cfg(feature = "anyhow")]
 #[test]
 fn test_option_ok_or_eyre() {
     let _ = eyre::set_hook(Box::new(|_e| {
--- a/eyre/tests/test_location.rs
+++ b/eyre/tests/test_location.rs
@@ -1,7 +1,5 @@
 use std::panic::Location;
 
-use eyre::WrapErr;
-
 struct LocationHandler {
     actual: Option<&'static str>,
     expected: &'static str,
@@ -46,6 +44,7 @@ fn test_wrap_err() {
         Box::new(LocationHandler::new(expected_location))
     }));
 
+    use eyre::WrapErr;
     let err = read_path("totally_fake_path")
         .wrap_err("oopsie")
         .unwrap_err();
@@ -75,6 +74,7 @@ fn test_wrap_err_with() {
         Box::new(LocationHandler::new(expected_location))
     }));
 
+    use eyre::WrapErr;
     let err = read_path("totally_fake_path")
         .wrap_err_with(|| "oopsie")
         .unwrap_err();
@@ -106,6 +106,7 @@ fn test_context() {
         Box::new(LocationHandler::new(expected_location))
     }));
 
+    use eyre::WrapErr;
     let err = read_path("totally_fake_path")
         .context("oopsie")
         .unwrap_err();
@@ -122,6 +123,7 @@ fn test_with_context() {
         Box::new(LocationHandler::new(expected_location))
     }));
 
+    use eyre::WrapErr;
     let err = read_path("totally_fake_path")
         .with_context(|| "oopsie")
         .unwrap_err();


ten3roberts:
I see, good catch.

I recently learnt that conditional trait methods are a backwards compatability nightmare (though WrapErr is sealed thankfully), as there was a recent wasmtime discussion regarding failure to compile because a trait implemented new methods another crate didn't implement under a certain feature flag combiniation.

A better option would be to move these to a supertrait or ContextCompat, and make trait implement for all errors to expose the same behavior as present, without footguns.

LeoniePhiline:
That would be an entirely different changeset, though, right?

ten3roberts:
Yes, just a proposition, as it would alleviate the confusion with two identically named methods from two different traits.

From what I see in the current source code `ContextCompat` is only implemented for `Option`, while `WrapErr` has the `context` method for errors.

If you are willing to do this, or want me to do this, just let me know.

It is a larger change, but would alleviate some of the confusion and import confusion for downstream users

LeoniePhiline:
My intention was to work on https://github.com/eyre-rs/eyre/issues/136 - I only discovered https://github.com/eyre-rs/eyre/issues/147 when starting to look into #136.

If you have time, I'd be glad if you could implement your proposition. I am not entirely sure if I fully see your vision in my mind.

eyre/tests/test_location.rs
```diff
@@ -83,7 +83,6 @@ fn test_wrap_err_with() {
     println!("{:?}", err);
 }
 
-#[cfg(feature = "anyhow")]
```
ten3roberts, line 86: 👍

ten3roberts:
Got my idea here: https://github.com/eyre-rs/eyre/pull/150

I feel I need to clarify that this discussion is more about [8522f02](https://github.com/eyre-rs/eyre/pull/148/commits/8522f02518330d4a0178a0e940c4a910bb9299e0) than the removal of the feature gate of the `OptionExt` trait.

Just doing what I can to clear out any confusion and make eyre as consistent as possible while we can.

In short, we move the `context*` methods to `ContextCompat` and implement it for Options and Results, to make all anyhow compatibility methods reside in one trait rather than two. This has the downside currently of `wrap_err` not being exposed for Options, but this just leans into the semantics of `wrap an error with another higher level error`, which does not make much sense for an `Option`.

When the anyhow feature drops from default features, errors and anyhow+options will be more self-contained and not lead to the same `wrap_err` in `ContextCompat` and `WrapErr`, and `context` from `ContextCompat` and `WrapErr` traits anymore.

Let me know what you think of this.

eyre/tests/test_location.rs
```diff
@@ -122,6 +122,7 @@ fn test_with_context() {
         Box::new(LocationHandler::new(expected_location))
     }));
 
+    use eyre::WrapErr;
```
ten3roberts, line 125: Screenshot 2024-01-24 at 17 30 21

Github decides to conveniently fold this code to make it appear as if it is in a different function.

This is what lead to some of my initial confusion in the #138 PR until I spotted the little code fold when scanning upwards for the function name 😅 

Anyhow, nice clarification on the traits, makes in-browser reviewing easier, and I am working on a PR to potentially fix this entirely.

ten3roberts (approved):

eyre/tests/test_location.rs
LeoniePhiline: image

"Expand all" can be helpful in diffs of rather uniform code, such as structurally repetitive tests.

thenorili (approved):
LGTM -- good cleanup

ten3roberts:
Thank you for your work Leonie.

Sorry for leaving this hanging, had a rough time lately, but I am back now :tada:

LeoniePhiline:
Please do not worry at all! Your wellbeing is always more important. <3

Issue #149, opened 2024-01-24
[ BREAKING ] `ContextCompat` contains identically named methods to `WrapErr` which can be confusing and a footgun


ten3roberts:
This can be confusing as you can't tell which method is being called.

It also leads to a backwards compatability hazard for conditional trait methods, though these are currently alleviated by virtue of the trait being sealed.

Additionally, `WrapErr` *also* exposes anyhow compatibility methods like `context`, which leads to the same confusion and footguns in the other direction.

This has lead to misunderstandings, poor reasonability, as well as mistakes in code where we called the wrong method.

Consider making anyhow compatibility consistent under one trait.

ten3roberts:
Related: #147 and https://github.com/eyre-rs/eyre/pull/138#discussion_r1432909562

PR #150, opened 2024-01-24
Make `ContextCompat` consistent and non-ambiguous with `WrapErr`

ten3roberts:
See: #149

Fixes: #149


thenorili (changes requested):
This looks like a good change to me, that ambiguity between option and result is detrimental and doesn't fit into our idea of what errors are.

It's a breaking change, so it'll need to wait for a major release.

Could you possibly write up a commit message to match? Other than that it looks fine to go on the experimental branch.

eyre/tests/test_location.rs
```diff
@@ -154,7 +158,7 @@ fn test_option_compat_wrap_err_with() {
     }));
 
     use eyre::ContextCompat;
-    let err = None::<()>.wrap_err_with(|| "oopsie").unwrap_err();
+    let err = None::<()>.with_context(|| "oopsie").unwrap_err();
```
ten3roberts, line 161: This correctly adheres to our model now as you can not "wrap an error" over an option.

You now either use the anyhow compatibility flag, or as the `AnyhowCompat` suggests the std methods like `ok_or_else` etc

ten3roberts:
@thenorili I've fixed up the commint message now and cleanup up the PR, let me know what you think :blush:

thenorili (approved):
a couple typos, lgtm

> The anyhow compatiblity methods (`context` and `with_context`) where
introduced as conditional required methods for the trait `WrapErr`, as
well as another separate trait `ContextCompat`, implemented for options
*and* results.

s/compatiblity/compatibility
s/where/were

thenorili (commented):
>This led to confusion as the two distinct trait 
s/trait/traits

thenorili (changes requested):
> This led to confusion as the two distinct trait had the same methods,
except one was implemented on results, and one of options and results.

suggested:

> This led to confusion as the two distinct traits had the same methods,
one was implemented on results and the other on both options and results.

---

> Further, this did not align with our error model, wherein errors are
wrapped with other errors in a chain now that options (which are no
errors and have no message) could be "wrapped"

s/no/not

suggested:

> Furthermore, wrapping options doesn't align with our error model where errors are wrapped around other errors; an error wrapping "None" should just be a single error.

thenorili (approved):

ten3roberts:
Thanks, will merge the CI fix before to get this to pass

eyre/tests/test_location.rs
```diff
@@ -117,12 +118,13 @@ fn test_context() {
 #[cfg(feature = "anyhow")]
 #[test]
 fn test_with_context() {
+    use eyre::ContextCompat;
```
LeoniePhiline, line 121: As above:

> For consistency with the other tests, this `use` could be moved to after the hook setup, just before _using_ `ContextCompat::with_context`.
> 
> Or the existing `use`s moved to the start of each test fn as well. 
> 
> Examples of `use` right before trait usage:
> 
> - https://github.com/eyre-rs/eyre/pull/150/files#diff-1ff47dac6cf55e34ff587968c5b1f1ec6b6ae39d2668a66ecba3633163a21fc5L142-R145 
> - https://github.com/eyre-rs/eyre/pull/150/files#diff-1ff47dac6cf55e34ff587968c5b1f1ec6b6ae39d2668a66ecba3633163a21fc5L157-R160

eyre/tests/test_location.rs
```diff
@@ -100,12 +100,13 @@ fn test_option_ok_or_eyre() {
 #[cfg(feature = "anyhow")]
 #[test]
 fn test_context() {
+    use eyre::ContextCompat;
```
LeoniePhiline, line 103: For consistency with the other tests, this `use` could be moved to after the hook setup, just before _using_ `ContextCompat::with_context`.

Or the existing `use`s moved to the start of each test fn as well. 

Examples of `use` right before trait usage:

- https://github.com/eyre-rs/eyre/pull/150/files#diff-1ff47dac6cf55e34ff587968c5b1f1ec6b6ae39d2668a66ecba3633163a21fc5L142-R145 
- https://github.com/eyre-rs/eyre/pull/150/files#diff-1ff47dac6cf55e34ff587968c5b1f1ec6b6ae39d2668a66ecba3633163a21fc5L157-R160

LeoniePhiline (commented):
Noticed some inconsistency in trait `use` placement within tests, where `use eyre::ContextCompat` was moved upwards from the original `use eyre::WrapErr` placement, while other `use eyre::ContextCompat` remain placed right before trait method invocation.

eyre/src/lib.rs
```diff
@@ -643,6 +645,7 @@ impl dyn EyreHandler {
         }
     }
 
+    /// Downcast the handler to a contcrete type `T`
```
thenorili, line 648: nit "concrete"

thenorili (approved):
LGTM

eyre/tests/test_location.rs
```diff
@@ -140,7 +142,7 @@ fn test_option_compat_wrap_err() {
     }));
 
     use eyre::ContextCompat;
-    let err = None::<()>.wrap_err("oopsie").unwrap_err();
+    let err = None::<()>.context("oopsie").unwrap_err();
```
LeoniePhiline, line 145: The encosing test `test_option_compat_wrap_err` might better be renamed to `test_option_compat_context`?

eyre/tests/test_location.rs
```diff
@@ -155,7 +157,7 @@ fn test_option_compat_wrap_err_with() {
     }));
 
     use eyre::ContextCompat;
-    let err = None::<()>.wrap_err_with(|| "oopsie").unwrap_err();
+    let err = None::<()>.with_context(|| "oopsie").unwrap_err();
```
LeoniePhiline, line 160: The encosing test `test_option_compat_wrap_err_with` might better be renamed to `test_option_compat_with_context`?

eyre/tests/test_location.rs
LeoniePhiline: `test_option_compat_context` and `test_option_compat_with_context` already exist and are (now) identical. The tests `test_option_compat_wrap_err` and `test_option_compat_wrap_err_with` are obsolete and should be deleted.

LeoniePhiline:
https://github.com/eyre-rs/eyre/pull/179 should fix my comments.

eyre/tests/test_location.rs
ten3roberts: And this is why consistent naming is so good, cause now that yields a duplicate definition error.

Thanks a lot for your suggestions 😊

SuperFluffy:
This caught me today. I would love to see this released soon. 🙏

PR #154, opened 2024-02-02
Document automated release process

ten3roberts:


CONTRIBUTING.md
```diff
@@ -173,9 +173,27 @@ both high-quality and timely.
...
 labeled and all resolved issues closed, feel free to message any Eyre Circle
-Member (currently @yaahc) for the triage role to help us keep things tidy. This
+Member (currently @yaahc and @ten3roberts) for the triage role to help us keep things tidy. This
 role only requires good faith and a basic understanding of our development
 process.
 
 [Discord]: https://discord.gg/z94RqmUTKB
 [^1]: Okay, I'll admit it, it's really just the Rust Project's CoC :sweat_smile:
+
+# Release Process
```
ten3roberts, line 183: We can always refine this, but we need something out now and ensure that releases go smoothly and don't drop commits

CONTRIBUTING.md
```diff
@@ -173,9 +173,27 @@ both high-quality and timely.
...
+
+The `master` branch contains all the latest changes and is considered unstable.
+
+For each bugfix and non-breaking change, the relevant commits are cherry-picked to relevant stable release branches,
+denoted by `releas-<major>.<minor>`. A new version is then published from each of the relevant stable release branches.
```
thenorili, line 188: s/releas/release

CONTRIBUTING.md
```diff
@@ -173,9 +173,27 @@ both high-quality and timely.
...
+
+Change logs are maintained on the `master` branch in the `CHANGELOG.md` file. Updating the changelog as part of a PR is
+highly encouraged, but not strictly required. The changelog *should not* be maintained on the stable release branches or
+in the release commits, as this leads to them not being visible on the `master` branch.
```
thenorili, line 192: Could you please explain in more detail what you mean by "The changelog *should not* be maintained on the stable release branches or in the release commits, as this leads to them not being visible on the `master` branch."?

It's more important for users to have a changelog for releases than to have a changelog for master.

An automated changelog for releases would be fine if we can get that working.

thenorili (commented):
Is the return to cargo-release fully automatic like the commit description claims? IIRC a fair bit of what we were doing manually was handled by cargo-release previously -- have we checked that against the manual procedure to make sure it hasn't rotted?

Issue #50, opened 2021-02-18
Dependency `once_cell v1.4.1` not compatible with `tokio v1.2.0` (which uses 1.5.2)


d4h0:
Hi,

I just started a project with `tokio`, and it turned out, the current versions of `eyre` and `tokio` are not compatible because different versions of `once_cell` are used.

Would it be possible for `eyre` to upgrade to the current version of `once_cell`?

d4h0:
Oh, never mind. I figured out how to get my code compile without changing `eyre`... 😆

In case someone has the same problem, and stumbled upon this via Google: I disabled the `tokio` features that depend on `once_cell` (`process` and `signal`).

It seems, many other crates still depend on `once_cell v1.4.1` (even crates that depend on `tokio`), so I will close this now. But fell free to re-open, if you think it makes sense.

yaahc:
I'm going to reopen this issue because I expect we'd still like to make sure our dependencies are as up to date as possible.

thenorili:
We're at v1.18 now.

PR #158, opened 2024-02-15
Update to once_cell v1.19.0

thenorili:
In the spirit of #50


thenorili:
Per discussion, 18 and 19 are compatible and we don't have a compelling reason for 1.18 not to be supported right now.

Issue #84, opened 2022-08-05
build.rs probe fails to propagate `--target` and does not use RUSTC_WRAPPER


RalfJung:
In tricky build situations, such as when bootstrapping rustc or for cross-compilation, it is crucial to set the right `--target` argument for rustc invocations. The build.rs script of eyre doesn't do that, and I think that's why I am now getting build failures in eyre inside a rustc bootstrap situation:
```
error[E0407]: method `backtrace` is not a member of trait `StdError`
   --> /home/r/.cargo/registry/src/github.com-1ecc6299db9ec823/eyre-0.6.8/src/context.rs:147:5
    |
147 | /     fn backtrace(&self) -> Option<&Backtrace> {
148 | |         self.error.backtrace()
149 | |     }
    | |_____^ not a member of trait `StdError`
```

Basically, I think eyre needs the same patch as https://github.com/dtolnay/anyhow/pull/249.
(FWIW autocfg gets all this right, but it seems a lot of crates re-implement the autocfg logic -- incorrectly.)

RalfJung:
There are also other issues with this build probe:
- it doesn't propagate CARGO_ENCODED_RUSTFLAGS, and hence will fail with `-Zallow-features` is used.
- it doesn't respect RUSTC_WRAPPER, though I am not sure if that actually causes trouble.

RalfJung:
Due to https://github.com/rust-lang/rust-analyzer/issues/12973 I would currently advise against honoring RUSTC_WRAPPER.

thenorili:
I'll fold this into my active work on the compile probe, thanks for the concern!

RalfJung:
Note that https://github.com/rust-lang/rust-analyzer/issues/12973 has been fixed by now. The anyhow probe is probably a good template to follow.

thenorili:
I'm already copying over their probe in the pr where i get backtrace working again : )

RalfJung:
eyre not honoring RUSTC_WRAPPER is now blocking improvements in Miri (https://github.com/rust-lang/miri/pull/3411). Miri would really like to rely on *all* rustc invocations going through the wrapper, it is very hard to support build scripts that bypass this.

autocfg has a very well-tested build probe logic. @thenorili I wonder if you would be open to taking autocfg as a dependency, if https://github.com/cuviper/autocfg/issues/24 would get resolved and you could access its build probe logic? @cuviper would you be willing to expose support for custom probes from autocfg, given a clear usecase like this one?

[...]

Issue #97, opened 2023-05-30
Backtrace is not available due to Error API change


ileixe:
I'm not sure this issue is tracking or not (or even intended), current build.rs script do not add `backtrace` feature as it failed to compile since Error does not have `backtrace()` API rather https://rust-lang.github.io/rfcs/3192-dyno.html.

I found this issue to compare anyhow and found that eyre does not emit valid backtrace like anyhow. The reason is that eyre do not set backtrace feature as I explained above. 

Anyhow has changed API (https://github.com/dtolnay/anyhow/commit/46d3d2c9f5260e74b0ff4c6262713f51b843e9de).

I modified to follow the latest API and it worked at least with minimal change. Can we as eyre support such change as well?

thenorili:
Thanks for this suggestion, I'll make these modifications and test it locally. Feel free to open a PR if you have it saved already, otherwise I'll get to it this week.

thenorili:
Ah yeah it looks like this has already changed, any::Demand doesn't exist.

https://github.com/dtolnay/anyhow/blob/master/build.rs

Anyhow has updated again it looks like. Mind you our backtrace module is much smaller than anyhow's, so I'm not sure that the minimal change you had in mind will work, but I'll give it a shot!

thenorili:
You're right that backtrace isn't working, this is important. I just implemented it for eyre. I'm going to write unit tests for this before I make a pull request. Thanks so much for bringing this to our attention! :D

ten3roberts:
:sweat_smile: Just saw your PR @thenorili, I accidentally fixed it too while working on another compile probe issue in #160.

Any chance we could merge that first and then look back into your PR which adds some more tests etc?

thenorili:
> 😅 Just saw your PR @thenorili, I accidentally fixed it too while working on another compile probe issue in #160.
> 
> Any chance we could merge that first and then look back into your PR which adds some more tests etc?

Yes, I think getting the fix in here before developing cohesive UI tests for backtrace is a good idea. I saw you've put it on main -- do you think this should be closed for now?

PR #124, opened 2023-11-20
Update to the new backtrace API

thenorili:
This addresses Issue #97. Pardon the half-baked commit message for now, I'll write it out more thoroughly once it's ready. I wrote a minimal test that produces error and panic backtraces, where I'm struggling is capturing this output and comparing it to a snippet. Comparing an error backtrace seems really difficult without intercepting stderr with something like Gag, which I'd prefer to avoid. Comparing the panic backtrace *seems* promising but the panic hook I added is only sending me the actual panic text -- ie panic!("foo") -> foo.

trying for an error backtrace like..

```
    fn low(frame: FailFrame) -> Result<(), Report> {
        let e: Report = eyre!("This program's frameness is suspect!");
        if let FailFrame::Low = frame {
            Err::<(), Report>(e).wrap_err("The low-level code has failed!")
            //            panic!("hello");
            //            ()
        } else {
            Ok(())
        }
    }

```

```
            Box::new(move |info| {
                let mut global_buffer = global_buffer.lock().unwrap();

                if let Some(s) = info.payload().downcast_ref::<&str>() {
                    println!("PANIC: {}", *s);
                    *global_buffer = (*s).to_string()
                } else {
                    panic!("help!");
                }
            })
        });

        panic::catch_unwind(|| {
            high(FailFrame::Low).unwrap();
        })
        .expect_err("Backtrace test did not panic.");
```

returns "panicked while processing panic", aka the payload downcast returned None.

PR #160, opened 2024-03-26
Fix compile probe and nightly backtraces

ten3roberts:
This addresses the invalid compile probe testing for a non-longer existing feature by updating it and eyre to use the `error_generic_member_access` features instead.

The report and errors have all been updated to accomodate this and the new backtrace provide API

Fixes: #84 and #97


ten3roberts:
I have to pin `trybuild` as it has added a dependency to `toml`  which only supports >1.70

ten3roberts:
I fixed it in https://github.com/eyre-rs/eyre/pull/161 and included the fix here

eyre/build.rs
```diff
@@ -26,59 +56,44 @@ const BACKTRACE_PROBE: &str = r#"
...
+    let probefile = Path::new(&out_dir).join("probe.rs");
+    fs::write(&probefile, probe).ok()?;
 
-fn main() {
-    match compile_probe(BACKTRACE_PROBE) {
-        Some(status) if status.success() => println!("cargo:rustc-cfg=backtrace"),
-        _ => {}
-    }
+    // Supports invoking rustc thrugh a wrapper
+    let mut cmd = if let Some(wrapper) = env::var_os("RUSTC_WRAPPER") {
```
RalfJung, line 73: It turns out there is also RUSTC_WORKSPACE_WRAPPER which should probably be honored as well.
[This](https://github.com/dtolnay/proc-macro2/blob/master/build.rs) is a reasonable template to start with.

thenorili:
It looks like this isn't quite finished. Have you considered working from this PR to fix nightly backtraces?

https://github.com/eyre-rs/eyre/pull/124

eyre/tests/generic_member_access.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    let err = fail()
+        .wrap_err("Failed to bake my favorite cupcake")
+        .unwrap_err();
+
+    let err: Box<dyn std::error::Error> = err.into();
+
+    assert!(
```
ten3roberts, line 55: Tests that eyre reports format a backtrace with `{:?}` and that it correctly contains the `fail` function.

If the backtrace was captured at the `wrap_err` conversion and not provided by the inner error it would not include this frame

eyre/tests/generic_member_access.rs
```diff
@@ -0,0 +1,73 @@
...
+        format!("{:?}", err).contains("generic_member_access::generic_member_access::fail"),
+        "should contain the source error backtrace"
+    );
+
+    assert_eq!(
```
ten3roberts, line 60: Tests the generic provide method

README.md
```diff
@@ -141,7 +144,7 @@ avoid using `eyre::Report` as your public error type.
 - Eyre works with any error type that has an impl of `std::error::Error`,
   including ones defined in your crate. We do not bundle a `derive(Error)` macro
   but you can write the impls yourself or use a standalone macro like
-  [thiserror].
+  [thiserror](https://github.com/dtolnay/thiserror).
```
LeoniePhiline, line 147: If resolving the link here, then https://github.com/eyre-rs/eyre/pull/160/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L198 can be removed.

eyre/src/wrapper.rs
thenorili, line 1: Not your code / out of scope: We're getting a couple dead code errors in this file with DisplayError and NoneError. I'm surprised they're triggered even though we marked them public. Why don't we tag these structs as `#[allow(dead_code)]`?

I am similarly distressed by the other warnings we get a dozen copies of down below. I thought the missing_doc_code_examples thing was resolved ages ago.

eyre/build.rs
thenorili, line 1: What was your motivation for rewriting this file?

README.md
```diff
@@ -178,6 +181,15 @@ No-std support was removed in 2020 in [commit 608a16a] due to unaddressed upstre
...
 https://github.com/eyre-rs/eyre/pull/29/commits/608a16aa2c2c27eca6c88001cc94c6973c18f1d5
 
+
+## Backtrace support
+
+The built in default handler has support for capturing backtrace using `rustc-1.65` or later.
+
+Backtraces are captured when an error is converted to an `eyre::Report` (such as using `?` or `eyre!`).
+
+If using the nightly toolchain, backtraces will also be captured and accessed from other errors using [error_generic_member_access](https://github.com/rust-lang/rfcs/pull/2895) if available.
```
thenorili, line 191: Per #121, some folks will continue to be unhappy about enforcing unstable features whenever we detect nightly. IIRC we discussed a while ago that we want to accomodate that use case. It doesn't make it any worse so I think it's fine to put that work off, but I thought it was worth mentioning.

thenorili (approved):
This looks okay to me to move forward with, sorry for the slowness!

eyre/src/wrapper.rs
RalfJung: > I'm surprised they're triggered even though we marked them public. 

`pub` in a private module can still be dead code.

thenorili: Yes, I understand that. I guess my surprise is that it's existing code that's had attention paid to its visibility and yet it's dead. I'll find the last time it wasn't dead code and figure its story out.

README.md
ten3roberts: Yeah, not wanting to break existing behavior in a "fix" PR, but rather make existing up to date nightly a tricky beast when it comes to libraries.

PR #161, opened 2024-03-26
Fix ci breaking due to new dependencies and rustc changes

ten3roberts:


ten3roberts:
Failing on windows due to an extra function call before main (in Rust) on windows only

color-eyre/tests/theme.rs
```diff
@@ -168,7 +170,7 @@ fn test_backwards_compatibility(target: String, file_name: &str) {
     fn normalize_backtrace(input: &str) -> String {
         input
             .lines()
-            .take_while(|v| !v.contains("core::panic"))
+            .take_while(|v| !v.contains("core::panic") && !v.contains("theme_test_helper::main"))
```
thenorili, line 173: This is ignoring a specific kind of panic from theme_test_helper, right?

thenorili (approved):
I gave my understanding of what you're doing here -- please correct me if I'm wrong, feel free to commit if not.

out of scope but we/I need to handle that huge block of warnings.

PR #163, opened 2024-03-29
 Fix ci breaking due to new dependencies and rustc changes (backport to master)

ten3roberts:
Applies the same fixes that was applied to `color-eyre-0.6` to master, namely the new update to `trybuild` that breaks our MSRV requirement, requiring us to pin the version, as the dependency only bumped the patch version for the updated MRSV.

It also fixes the theme test breaking on windows due to new *before-main* machinery on windows affecting the backtrace. This is fixed by filtering out the functions before and including main(due to no take_until iter adapter in std), but still includes the panicking function call.

There are currently three branches that need/needed this fix

- master: containing latest breaking code for 1.0
- color-eyre-0.6
- release-0.6 (does not include the theme test change as color-eyre is not present there)

### DIFF
--- a/color-eyre/tests/theme.rs
+++ b/color-eyre/tests/theme.rs
@@ -124,6 +124,8 @@ fn test_panic_backwards_compatibility() {
         .args(["run", "--example", "theme_test_helper"])
         .arg("--no-default-features")
         .args(&features)
+        .env("RUSTFLAGS", "-Awarnings")
+        .env("CARGO_FUTURE_INCOMPAT_REPORT_FREQUENCY", "never")
         .output()
         .expect("failed to execute process");
     let target = String::from_utf8(output.stderr).expect("failed to convert output to `String`");
@@ -168,7 +170,7 @@ fn test_backwards_compatibility(target: String, file_name: &str) {
     fn normalize_backtrace(input: &str) -> String {
         input
             .lines()
-            .take_while(|v| !v.contains("core::panic"))
+            .take_while(|v| !v.contains("core::panic") && !v.contains("theme_test_helper::main"))
             .collect::<Vec<_>>()
             .join("\n")
     }
--- a/eyre/Cargo.toml
+++ b/eyre/Cargo.toml
@@ -27,7 +27,7 @@ pyo3 = { version = "0.20", optional = true, default-features = false }
 futures = { version = "0.3", default-features = false }
 rustversion = "1.0"
 thiserror = "1.0"
-trybuild = { version = "1.0.19", features = ["diff"] }
+trybuild = { version = "=1.0.83", features = ["diff"] }
 backtrace = "0.3.46"
 anyhow = "1.0.28"
 syn = { version = "2.0", features = ["full"] }


eyre/Cargo.toml
```diff
@@ -27,7 +27,7 @@ pyo3 = { version = "0.20", optional = true, default-features = false }
 futures = { version = "0.3", default-features = false }
 rustversion = "1.0"
 thiserror = "1.0"
-trybuild = { version = "1.0.19", features = ["diff"] }
+trybuild = { version = "=1.0.83", features = ["diff"] }
```
ten3roberts, line 30: Needed for #150 to pass

color-eyre/tests/theme.rs
```diff
@@ -124,6 +124,8 @@ fn test_panic_backwards_compatibility() {
         .args(["run", "--example", "theme_test_helper"])
         .arg("--no-default-features")
         .args(&features)
+        .env("RUSTFLAGS", "-Awarnings")
+        .env("CARGO_FUTURE_INCOMPAT_REPORT_FREQUENCY", "never")
```
ten3roberts, line 128: This removes cargo printing an extra line on windows warning about a future incompatability in nom.

The cargo output is included in the panic test case as it invokes cargo run and looks at the output.

Since the report is colored with ansi-codes, it gets included in the test check

ten3roberts:
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/examples/theme_test_helper`
[31mThe application panicked (crashed).[0m
Message:  [36m<non string panic payload>[0m
Location: [35mexamples/theme_test_helper.rs[0m:[35m37[0m

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: [91mtheme_test_helper[0m[91m::[0m[91mget_error[0m with [96m[3mmsg[0m[2m=[0m"test"[0m
      at [35mexamples/theme_test_helper.rs[0m:[35m34[0m

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
  [96m                              ⋮ 6 frames hidden ⋮                               [0m
   7: [32mstd::panic::panic_any[0m[90m::hd76a7f826307234c[0m
      at [35m/rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/panic.rs[0m:[35m57[0m
   8: [91mtheme_test_helper::main[0m[90m::h767d3fd6c45048c8[0m
      at [35m/home/jlusby/git/yaahc/color-eyre/examples/theme_test_helper.rs[0m:[35m37[0m
   9: [32mcore::ops::function::FnOnce::call_once[0m[90m::hc5a1cd4127189dad[0m
      at [35m/rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/ops/function.rs[0m:[35m227[0m
  [96m                              ⋮ 15 frames hidden ⋮                              [0m

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.
```

This is what we are checking against, note how it included the cargo output as well as the running output.

This has been an issue in the past wrt existing warnings, which was fixed by fixing the warnings.

*This* warning will fix itself in nom and we can not currently do anything about the warning

thenorili (approved):
This seems completely satisfactory.

PR #167, opened 2024-04-17
Bump workspace to owo-colors 4.0

workingjubilee:
Requires adding a path dependency in color-eyre on color-spantrace so that the versions agree. Fortunately we can do that now!


color-eyre/Cargo.toml
```diff
@@ -23,7 +23,7 @@ tracing-error = { version = "0.2.0", optional = true }
 backtrace = { version = "0.3.48", features = ["gimli-symbolize"] }
 indenter = { workspace = true }
 owo-colors = { workspace = true }
-color-spantrace = { version = "0.2", optional = true }
+color-spantrace = { version = "0.2", path = "../color-spantrace", optional = true }
```
pksunkara, line 26: Unless I am mistaken, We don't need a path dependency to get the versions to agree. What error were you getting?

I am planning to make this a proper workspace once the crates are merged too, but I would prefer if we don't add path dependencies until that point so that I could use workspace path dependencies.

.gitignore
```diff
@@ -1,4 +1,3 @@
 /target
 **/*.rs.bk
-Cargo.lock
```
pksunkara, line 3: What's the reason for committing the lockfile?

workingjubilee: https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

color-eyre/Cargo.toml
workingjubilee: @pksunkara You are indeed mistaken.

```toml
color-spantrace = { version = "0.2", path = "../color-spantrace", optional = true }
```
meant that it was resolving to the color-spantrace on crates.io, which depends on owo-colors 3.5, which is not compatible with 4.0, which meant that the types inside color-eyre were disagreeing on the version, which was preventing the conversion in the relevant `From` impl.

workingjubilee: Specifically this code does not compile:
https://github.com/eyre-rs/eyre/blob/7a5c32acd7a4a6139448b7900a6787f70b5b69fb/color-eyre/src/config.rs#L768-L778

pksunkara: I forgot we haven't published this. Thanks.

pksunkara:
Windows tests are failing.

PR #170, opened 2024-04-24
fix(docs): enclose inline code in backticks

### LINKED ISSUE #169: docs: enclosing in backticks
Reading https://docs.rs/eyre/latest/eyre/struct.Report.html#display-representations I found that some items should better be enclosed in backticks.

image

I'll happily fix that, I merely want to ask in advance:

When enclosing parts of strings in backticks, do you then prefer **backticks + quotes**, or **only backticks**?

I personally prefer backticks + quotes, to signify _"this is code, but it's a string literal"_.

**Here an example (look out for the curly braces):**

1. **backticks + quotes**: When you print an error object using `"{}"` or `to_string()`, only the outermost underlying error is printed, not any of the lower level causes. This is exactly as if you had called the `Display` `impl` of the error from which you constructed your `eyre::Report`. [...] To print causes as well using eyre’s default formatting of causes, use the alternate selector `"{:#}"`.
2. **only backticks**: When you print an error object using `{}` or `to_string()`, only the outermost underlying error is printed, not any of the lower level causes. This is exactly as if you had called the `Display` `impl` of the error from which you constructed your `eyre::Report`. [...] To print causes as well using eyre’s default formatting of causes, use the alternate selector `{:#}`.

LeoniePhiline:
Fixes #169

Besides enclosing inline code in backticks, I also made two tiny documentation improvements - please let me know if you want them undone:

- In the `Report::msg` docs, the user is referred to `Report::new`. I changed this from plain inline code into a link.
- I expanded _"`Display` impl"_ to _"`Display` implementation"_, since `impl` is a Rust token, but 
  - _"`Display` `impl`"_ looked silly, and 
  - _"`impl Display`"_ disturbed the flow of reading.

### DIFF
--- a/eyre/src/error.rs
+++ b/eyre/src/error.rs
@@ -27,7 +27,7 @@ impl Report {
 
     /// Create a new error object from a printable error message.
     ///
-    /// If the argument implements std::error::Error, prefer `Report::new`
+    /// If the argument implements `std::error::Error`, prefer [`Report::new`]
     /// instead which preserves the underlying error's cause chain and
     /// backtrace. If the argument may or may not implement std::error::Error
     /// now or in the future, use `eyre!(err)` which handles either way
--- a/eyre/src/lib.rs
+++ b/eyre/src/lib.rs
@@ -417,22 +417,22 @@ pub use WrapErr as Context;
 ///
 /// # Display representations
 ///
-/// When you print an error object using "{}" or to_string(), only the outermost underlying error
-/// is printed, not any of the lower level causes. This is exactly as if you had called the Display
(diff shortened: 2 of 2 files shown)


LeoniePhiline:
The CI failures confuse me. They seem unrelated?

pksunkara (approved):

yaahc:
@LeoniePhiline I think we'll land 163 then you can go ahead and rebase on master

yaahc:
turns out I was able to do the rebase myself easily ^_^

LeoniePhiline:
Thank you! ☺️

PR #172, opened 2024-05-15
Fix overly strict clippy lint

ten3roberts:
Deny style, suspicious and complexity categories, rather than all warnings

If we find issues that could have been caught by a clippy lint we can add that lint on as *deny* too, as for now, they still appear as warnings but won't prevent CI from running.

This most notably fixes random CI breakages when clippy/rust updates and introduces new warn level lints

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,8 +135,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets --all-features -- -D warnings
-
+          args: --all-targets --all-features -- -D clippy::style -D clippy::suspiscious -D clippy::complexity
   miri:
     name: Miri
     runs-on: ubuntu-latest


.github/workflows/ci.yml
```diff
@@ -135,8 +135,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets --all-features -- -D warnings
-
+          args: --all-targets --all-features -- -D clippy::style -D clippy::suspiscious -D clippy::complexity
```
pksunkara, line 138: Instead of doing it like this, it might be worth changing the `stable` to an MSRV.

ten3roberts: That makes sense, pinning the clippy version

Only danger with that is that we may not update it for a while and not get new "good to deny by default" lints.

This primarily solves the overly pedantic lints, while still giving us all the new ones as warnings.

Both are good

Maybe a combination?

Sticking to latest also makes it easier to run clippy locally, as you don't have to track which version, especially for new contributors who aren't aware of it, not understanding why there are different results.

But that is just a minor thing.

PR #178, opened 2024-06-21
Support converting between `anyhow::Error` and `eyre::Report` automatically with `?`

ten3roberts:


Cargo.toml
```diff
@@ -29,6 +29,7 @@ pyo3 = { version = "0.13", default-features = false, features = ["auto-initializ
 indenter = "0.3.0"
 once_cell = "1.4.0"
 pyo3 = { version = "0.13", optional = true, default-features = false }
+anyhow = "1.0"
```
pksunkara, line 32: Let's put this behind a feature?

Issue #135, opened 2023-12-20
[breaking] remove alias exports `DefaultContext` and `EyreContext`


LeoniePhiline:
The following aliasing exports are unused and can be removed:

```rust
#[cfg(feature = "anyhow")]
#[doc(hidden)]
pub use DefaultHandler as DefaultContext;
#[cfg(feature = "anyhow")]
#[doc(hidden)]
pub use EyreHandler as EyreContext;
```

The aliases might be imported downstream, making this a breaking change.

LeoniePhiline:
Status:#181 is ready for review and merge.

PR #181, opened 2024-06-28
chore!: remove alias exports `DefaultContext` and `EyreContext`

LeoniePhiline:
`DefaultContext` and `EyreContext` exist as alias exports enabled by the `anyhow` feature.

However, they are unused within `eyre`; neither are handlers of such names exposed by `anyhow`.

This changeset removes both above mentioned handler aliases.

Closes [#135][135].

[135]: https://github.com/eyre-rs/eyre/issues/135

BREAKING CHANGE: The alias exports are undocumented but nevertheless part of the crate's public API.

### DIFF
--- a/eyre/src/lib.rs
+++ b/eyre/src/lib.rs
@@ -389,12 +389,6 @@ pub use eyre as format_err;
 pub use eyre as anyhow;
 use once_cell::sync::OnceCell;
 use ptr::OwnedPtr;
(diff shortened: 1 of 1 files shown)


LeoniePhiline:
Rebased.

LeoniePhiline:
Thanks, @pksunkara! :)

PR #175, opened 2024-05-31
Bump backtrace-rs

kristof-mattei:
0.3.59 is the first version of backtrace-rs that doesn't use `gimli-symbolize`: https://github.com/rust-lang/backtrace-rs/commit/5fc4f79674206c2041b9ffeec6e3891c73510147

While the feature is still there, it is unused and later on removed in https://github.com/rust-lang/backtrace-rs/commit/4cbe6b69eaccf23297b51f9584dd8a7d5e12ea95. Original [PR #615](https://github.com/rust-lang/backtrace-rs/pull/615)

Closes https://github.com/eyre-rs/eyre/issues/174

PR #185, opened 2024-07-05
Use a recent backtrace

workingjubilee:
Drop the gimli-symbolize feature requirement because it does nothing, since a version from a long time ago.


workingjubilee:
Oh, I didn't notice #175

PR #188, opened 2024-07-31
feat(color-eyre): add default `export-owo-colors` feature

0xAdk:
Reexporting `owo-colors` causes rust-analzyer's code completion to fill with many auto-use suggestions for the `owo-colors::OwoColorize` trait.

This feature allows users to opt-out of reexporting `owo-colors`.

see https://github.com/eyre-rs/color-eyre/issues/109


0xAdk:
Sorry since I don't use features often this didn't occur to me. Would this be a breaking change for anyone with `default-features = false`? Maybe a non-default `no-export-owo-colors` (I'm sure there is a better name) feature would be better.

HactarCE:
Features are meant to be additive. Consider two crates, where the first one depends on those exports and the second one doesn't. If the second crate specifies `no-export-owo-colors`, then combining both crates into one project would cause the first crate to not compile because the `no-export-owo-colors` feature is active.

On the other hand, an `export-owo-colors` feature would work. If any crate in the dependency tree depends on it, it will be included in the final build.

Currently there are no default features, so there's no reason someone would have `default_features = false` for `color_owo`. It may still count as a breaking change, but it is unlikely to break code in practice.

I think `export-owo-colors` as a default feature is definitely the way to go.

0xAdk:
ok the feature is now additive.

> Currently there are no default features, so there's no reason someone would have `default_features = false` for `color_owo`

I'm not sure what you mean. There is both `track-caller` and `capture-spantrace` already enabled by default.

HactarCE:
Ah, my bad. I guess I looked in the wrong place for feature documentation

ten3roberts:
>  I think export-owo-colors is definitely the way to go. For minimal breakage, add it as a default feature (although I think most users of this crate will not want it enabled).

Agreed, 

Personally, I rarely export dependency crates unless they are *extensively* used in the external API, as the user can easily bring their own, rather than having two ways to import it. This is made somewhat easier with RA recent checks for `[doc_hidden]` attrs that prevents considering them for autoimport, but for many, it is not beneficial to always have access to, and accidentally auto-import `red` from `color_eyre::owo_colors`

PR #192, opened 2024-08-13
add support to specify verbosity in HookBuilder

TheTollingBell:
In situations where setting the environment is [unsafe](https://doc.rust-lang.org/std/env/fn.set_var.html#safety) it would be more convenient to manually specify the verbosity without using the environment at all.

### DIFF
--- a/color-eyre/src/config.rs
+++ b/color-eyre/src/config.rs
@@ -219,9 +219,9 @@ impl<'a> fmt::Display for StyledFrame<'a> {
         )?;
 
         let v = if std::thread::panicking() {
-            panic_verbosity()
+            panic_verbosity(None)
         } else {
-            lib_verbosity()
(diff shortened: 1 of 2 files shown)


yaahc:
Looks good, I'm happy to merge this with the addition of some tests. Particularly what happens when conflicting verbosity is set via both the hook builder and the env.

TheTollingBell:
Sorry for being incommunicado for the past few days, I'll go ahead and fix Clippy failing. As it's configured right now, verbosity that is specified in HookBuilder overrides environment, though it is possible to write another function that would instead have environment override specified verbosity.

TheTollingBell:
Never mind about the Clippy, all the warnings are not from changes from this pull request.

yaahc:
clippy warnings should be fixed now

PR #197, opened 2024-08-22
chore(color-eyre): Set color-eyre dependency on eyre via path.

humb1t:
Good day, dear eyre contributors.

I really like to use eyre in all my projects. Right now I use master to stay on 1.0 version.
Unfortunately when I want to start using color-eyre the same way it conflicts with eyre, because it uses 0.6.1 version.
What this PR addresses is to temporary set color-eyre -> eyre dependency via path, not a version, until 1.0 will be released.

If it's not something you want to have in the codebase, just please let me know, I will then switch to my fork and keep it sync with the main repo. Thank you in advance, you are awesome! 💜

### DIFF
--- a/color-eyre/Cargo.toml
+++ b/color-eyre/Cargo.toml
@@ -18,7 +18,7 @@ issue-url = ["url"]
 track-caller = []
 
 [dependencies]
-eyre = "0.6.1"
+eyre = { version = "1.0.0", path = "../eyre" }
 tracing-error = { version = "0.2.0", optional = true }
 backtrace = { version = "0.3.59" }
 indenter = { workspace = true }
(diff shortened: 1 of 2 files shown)


color-eyre/Cargo.toml
```diff
@@ -18,7 +18,7 @@ issue-url = ["url"]
 track-caller = []
 
 [dependencies]
-eyre = "0.6.1"
+eyre = { path = "../eyre" }
```
pksunkara, line 21: Please have a version too in here. That way, we get best of both worlds.

humb1t: Thank you @pksunkara — I added version to this dependency description as "1.0.0", is it okay?

pksunkara: Yes, but please reset the lockfile to remove unneeded upgrades. And then run `cargo update` to update eyre related changes.

humb1t: That's done @pksunkara

PR #208, opened 2024-11-07
feature request: Allow attaching Reports to Reports

### LINKED ISSUE #198: Feature request: Allow attaching Reports to Reports
While `with_error` already allows attaching errors, a `color_eyre::Report` would need to be coerced into a `StdError`, losing `HelpInfo` sections, and probably also `tracing` data in the process.

eyre-rs/color-eyre#125 would have implemented the option to attach proper `color_eyre::Report`s, which would allow for something like this:

```rust
let mut errors = Vec::new();
for entry in walker {
    if let Err(err) = do_stuff(entry) {
        errors.push(err);
        if params.fail_fast {
            break;
        }
    }
}

if !errors.is_empty() {
    let mut err = eyre::eyre!("Failed to do_stuff with {} errors", errors.len());

    for e in errors {
        err = err.report(e);
    }

    return Err(err);
}
```

This would render nicely, with the full details for all errors, including helpful sections like from `wrap_err` or `with_suggestion`.

afmireski:
Implemented a way to properly link additional `color_eyre::Report`s to a Report: `report` and `with_report`.

- [x] All existing tests passed
- [x] Documentation for new methods added
- [x] Usage examples provided

closes #198


afmireski:
Accepted suggestions from `cargo fmt` and `cargo clippy` to fix the CI errors related to `rustfmt` and `clippy`.
- [x] All existing tests passed

afmireski:
Regarding the `macOS-latest` CI failure, I believe updating the pyo3 crate to a more recent version that supports Python 3.13 might resolve the issue.
image

Since version 22.0, pyo3 seems to be adding support for Python 3.13.
However, I consider upgrading package versions a critical decision that shouldn't be made alone.

My proposal is to update the eyre Cargo.toml as follows:
```toml
[dependencies]
...
pyo3 = { version = "23.1", optional = true, default-features = false }
...

[dev-dependencies]
...
pyo3 = { version = "0.23.1", default-features = false, features = ["auto-initialize"] }
```
@yaahc, would this be a valid approach? I'm open to any suggestions or feedback.

yaahc:
tenatively that sounds fine, the only concern would be whats the MSRV for that version of pyo3 since we'd inherit it

afmireski:
Looking at the release history of the `pyo3` crate, the current version being used (`0.20`) has a MSRV of `1.56.0`. Starting from version `0.22.0`, the MSRV was raised to `1.63.0`.
Is it okay if I proceed with the change?

- [crate.io version history](https://crates.io/crates/pyo3/versions)
- [Release 0.22.0 Notes](https://github.com/PyO3/pyo3/releases/tag/v0.22.0)

afmireski:
Bump `pyo3` crate version to `0.23.1`.
- [x] All existing tests passed

**Observations:**
- The MSRV for Eyre remains `1.65.0` since `pyo3` `0.23.1` has an MSRV of `1.63.0`.
- [`test_pyo3.rs`](https://github.com/eyre-rs/eyre/pull/208/files#diff-1e9d522e92834e5d396c1f36d4829f0e82c1ea92976dc1a01da40beb7b5aafab) was refactored to fix some compiler errors caused by the `pyo3` version upgrade.

afmireski:
Hi, @yaahc I hope you're doing well! When you have some time, could you kindly take a look at my PR? Let me know if anything needs adjustment. Thanks!

color-eyre/src/section/help.rs
```diff
@@ -289,6 +327,20 @@ impl Display for HelpInfo {
...
+                let chain_errors = report.chain();
+                write!(f, "Report:")?;
+                for (n, error) in chain_errors.enumerate() {
+                    writeln!(f)?;
+                    write!(
+                        indented(f).ind(n),
+                        "{}",
+                        error.style(theme.help_info_report)
+                    )?;
+                }
```
vultix, line 342: If you have any backtrace or spantrace information in this report, it won't be printed. That feels like a motivating use case for this feature

afmireski: Hello @vultix , thank you very much for the feedback!

If it's not too much trouble, could you explain in more detail what seems to be wrong?

When running the examples I added, the backtrace information appears to be displayed correctly, very similar to how it is shown when running the error examples:
image

I would really appreciate any clarification you can provide.

sclu1034: The motivation in #198 (by me) was "render [...] including helpful sections like from `wrap_err` or `with_suggestion`".

Your screenshot does show one backtrace. It's for the parent error, which is regular behavior, unaffected by this PR.
But there should be three backtraces shown. The two "Report" sections at the bottom, the actual new feature, only show an error message, when they should show all their `HelpInfo` sections. That's what would set this apart from the existing `with_error`.

Rolv-Apneseth:
This would be a really nice feature to have, but I agree that having the backtrace for _all_ reports is quite important. Does the PR need any kind of help to get there @afmireski?

LeoniePhiline:
The removal of the `pyo3` feature in https://github.com/eyre-rs/eyre/pull/278 obsoletes some of the above discussion and will slightly simplify this PR.

PR #218, opened 2025-02-03
switch from `once_cell` to `std::sync::OnceLock`

jakobhellermann:
One less dependency is always nice to have, and [OnceLock](https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html) has been stable since 1.70.


Buzzec:
The MSRV must be updated for this.

pksunkara:
We are planning to increase the MSRV soon. Can you please update the PR?

PR #225, opened 2025-05-01
Update backtrace

LegNeato:
This is the last backtrace release that supports 1.65 MSRV. Through many layers miri uses this crate and it is pulling in an old version of `object` (0.32) where all of the `rustc` crates are on object 0.36.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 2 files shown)


LegNeato:
@yaahc if you are cutting a release I would love to get this in as I am trying to clean up some deps in `rustc`.

LegNeato:
Actually, this is not needed as long as a new release is pushed.

PR #238, opened 2025-05-22
Apply colors only when they are supported

### LINKED ISSUE #236: NO_COLOR support
Bump https://github.com/eyre-rs/color-eyre/issues/99.

> I love color-eyre when developing locally; it is so incredibly helpful! But the colors are a problem when deploying to prod where, for example, Cloudwatch logs for AWS Lambda functions just include raw ANSI escape sequences. I'm working around that right now by using this code in my setup function:
> 
>         if std::env::var("NO_COLOR") == Err(std::env::VarError::NotPresent) {
>             color_eyre::install()?;
>         } else {
>             color_eyre::config::HookBuilder::new()
>                 .theme(color_eyre::config::Theme::new())
>                 .install()?;
>         }
> and making sure our prod deployments include NO_COLOR=1.
> 
> I'd love to see either:
> 
> a mention of this particular pattern in the docs (I searched and couldn't find anything), or
> a change to HookBuilder::blank() selecting either the default dark theme or the blank theme based on the presence of NO_COLOR
> I'm sure you've already seen this, but I found https://no-color.org/ helpful
> 
> I'm sorry I don't have the time right now to write up a draft PR for you.

### LINKED ISSUE #237: Redirecting stderr to file doesn't remove special ANSI characters.
### Minimal reproducible example

Run this to see the MRE:

```shell
cargo new mre
cd mre
cargo add color-eyre
(
    echo 'fn main() {'
    echo '    color_eyre::install().unwrap();'
    echo '    panic!("oh no!");'
    echo '}'
) > src/main.rs
cargo run 2> oh_no.txt
nano oh_no.log
```

### Contents of resulting `oh_no.txt` file

Image

### Question

Is it possible to configure color_eyre/owo_colors to avoid printing special characters when piping or redirecting the output? It really bugs me :(

nikableh:
Closes #236.
Closes #237.

This PR wraps all calls to `owo_colors::OwoColorize::style` with function that calls `anstream::AutoStream::choice` and checks whenever colors are available for the current stderr.

### Conserns

The stderr here is hardcoded it that somewhat bothers me. It is possible for that user of eyre to configure printing all of the error messages into the stderr? Because if eyre ever prints color to the stdout and then the user redirects stdout to the file, [this check](https://github.com/eyre-rs/eyre/blob/99576ecac083080e9668472b2d7b18e301ec2ad0/color-spantrace/src/lib.rs#L392) would detect that stderr is still a terminal and allow ANSI symbols to be printed to the file. 

I am also wondering if duplicating `style_if_possible` in `color-spantrace` and `color-eyre` is okay. Should I move it somewhere else?

### Tests

I've done some quick testing by running examples and this seems to work well. I don't see any colors in the terminal and there are no special symbols _from eyre_ when redirecting to the file.

Here is a script to run all color-eyre examples and print the output to the file. Mind that `RUSTFLAGS=-Awarnings` will force everything to rebuild. But that saves you from scrolling past all the deprecation warnings (fixed here btw — https://github.com/eyre-rs/eyre/pull/235).

```bash
#!/bin/env bash

TEST_OUTPUT_DIR=test-no-color-examples
EXAMPLES='custom_filter custom_section debug_perf github_issue multiple_errors panic_compose panic_hook theme_test_helper theme color-eyre-usage'

rm -r $TEST_OUTPUT_DIR || true
mkdir -p $TEST_OUTPUT_DIR

for example in $EXAMPLES; do
    RUSTFLAGS=-Awarnings cargo run --example $example &>> $TEST_OUTPUT_DIR/$example
done
```

Might be nice to write some unit tests, but I'd prefer if someone else did it.

### About tracing_subscriber

The only ANSI symbols I see are the ones coming from tracing_subscriber ([color-eyre/examples/theme_test_helper.rs:35](https://github.com/eyre-rs/eyre/blob/master/color-eyre/examples/theme_test_helper.rs#L35)). I did a little digging and found that removing this layer [color-eyre/examples/theme_test_helper.rs:49](https://github.com/eyre-rs/eyre/blob/master/color-eyre/examples/theme_test_helper.rs#L49) from registry fixes the issue.

The colors are still there, in the terminal, but not in the file. The only thing that changed visually is that the text is no longer cursive. Although I don't quite understand what this layer does and why is it needed. The [documentation](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.with_target) is not clear to me.

I think we should just remove this layer from every example to avoid teaching people this flawed behaviour. Let's leave it up to them to enable this layer. This way it won't be eyre's problem, but tracing_subscriber's one.

### Rant about owo_colors

To be fair, this all feels like a huge blunder from owo_colors. I feel like it is it who should be responsible for detecting colors. They already have a `supports-color` feature. But enabling it only gives you a method that you should apply everywhere yourself. I can understand why this feature isn't turned on by default, sure. But once I enable it I kind of expect everything to work out of the box. What if you have thousands of lines of code where colors are added? Do you wrap everything with `if_supports_color`??? This is nonsense.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ indenter = "0.3.0"
 once_cell = "1.18.0"
 owo-colors = "4.0"
 autocfg = "1.0"
+anstream = "0.6.15"
 
 [profile.dev.package.backtrace]
 opt-level = 3
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ include = ["README.md", "LICENSE-APACHE", "LICENSE-MIT", "build.rs", "Cargo.toml
 indenter = "0.3.0"
 owo-colors = "4.3"
 autocfg = "1.0"
+anstream = "0.6.15"
 
 [profile.dev.package.backtrace]
 opt-level = 3
--- a/color-eyre/examples/theme.rs
+++ b/color-eyre/examples/theme.rs
@@ -1,12 +1,13 @@
-use color_eyre::{config::Theme, eyre::Report, owo_colors::style, Section};
+use color_eyre::{config::Theme, eyre::Report, Section};
+use owo_colors::Style;
 
 /// To experiment with theme values, edit `theme()` below and execute `cargo run --example theme`
 fn theme() -> Theme {
     Theme::dark()
         // ^ use `new` to derive from a blank theme, or `light` to derive from a light theme.
         // Now configure your theme (see the docs for all options):
-        .line_number(style().blue())
-        .help_info_suggestion(style().red())
+        .line_number(Style::new().blue())
+        .help_info_suggestion(Style::new().red())
 }
 
 #[derive(Debug, thiserror::Error)]
--- a/color-eyre/examples/theme.rs
+++ b/color-eyre/examples/theme.rs
@@ -1,12 +1,13 @@
-use color_eyre::{Section, config::Theme, eyre::Report, owo_colors::style};
+use color_eyre::{Section, config::Theme, eyre::Report};
+use owo_colors::Style;
 
 /// To experiment with theme values, edit `theme()` below and execute `cargo run --example theme`
 fn theme() -> Theme {
     Theme::dark()
         // ^ use `new` to derive from a blank theme, or `light` to derive from a light theme.
         // Now configure your theme (see the docs for all options):
-        .line_number(style().blue())
-        .help_info_suggestion(style().red())
+        .line_number(Style::new().blue())
+        .help_info_suggestion(Style::new().red())
 }
 
 #[derive(Debug, thiserror::Error)]
--- a/color-eyre/src/lib.rs
+++ b/color-eyre/src/lib.rs
@@ -364,6 +364,7 @@
 
 use std::sync::Arc;
 
+use anstream::AutoStream;
 use backtrace::Backtrace;
 pub use eyre;
 #[doc(hidden)]
@@ -371,6 +372,8 @@ pub use eyre::Report;
 #[doc(hidden)]
 pub use eyre::Result;
 pub use owo_colors;
+use owo_colors::OwoColorize;
+use owo_colors::Style;
 use section::help::HelpInfo;
 #[doc(hidden)]
 pub use section::Section as Help;
@@ -461,4 +464,18 @@ pub enum ErrorKind<'a> {
 /// ```
 pub fn install() -> Result<(), crate::eyre::Report> {
     config::HookBuilder::default().install()
+}
+
+/// Apply owo_colors style if possible. Returns a string with/without ANSI
+/// escape symbols.
+fn style_if_possible<S>(str: S, style: Style) -> String
+where
+    S: ToString + std::fmt::Display,
+{
+    let color_choice = AutoStream::choice(&std::io::stderr());
+
+    match color_choice {
+        anstream::ColorChoice::Never => str.to_string(),
+        _ => str.style(style).to_string(),
+    }
 }
--- a/color-eyre/src/lib.rs
+++ b/color-eyre/src/lib.rs
@@ -468,7 +468,7 @@ pub fn install() -> Result<(), crate::eyre::Report> {
 
 /// Apply owo_colors style if possible. Returns a string with/without ANSI
 /// escape symbols.
-fn style_if_possible<S>(str: S, style: Style) -> String
+fn style<S>(str: S, style: Style) -> String
 where
     S: ToString + std::fmt::Display,
 {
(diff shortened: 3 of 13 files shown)


pksunkara (changes requested):
Ideally, we can just call the `style_if_possible` function as `style`. From a developer's perspective, they don't care whether that function is styling or not, depending on the env vars.

pksunkara (changes requested):
Can you please do the following changes:

1. Replace all `owo_colors::style()` usage with `owo_colors::Style::new()` (needed for 2nd change)
2. Rename all `style_if_possible` to `style`.

epage:
`anstream` is a larger dependency for just determining color choice and has specific logic dealing with its auto-adapting to old versions of Windows.

You could pull out the choice logic, calling `anstyle_query` directly, and adapt to the needs here.

Or eyre could do something like clap and be neutral to this and assume the caller is using `anstream` with ways of stripping the color.

pksunkara (approved):
Looks good. I will merge once you can get CI passing.

aditeyapatakoti (approved):
nice edits

pksunkara:
@synalice Can you please fix CI?

pksunkara:
Still failing. Can you please try all the commands from CI locally first? Using github actions takes a lot of time since a maintainer needs to approve every time.

pksunkara (changes requested):
Looking at the failing tests, you need to enable colors in tests so that they can be tested.

svenstaro:
@nikableh Hey, thanks for your work on this! Any chance you can get this finished up so that my logs are cleaner from now on? :)

Cargo.toml
```diff
@@ -20,6 +20,7 @@ include = ["README.md", "LICENSE-APACHE", "LICENSE-MIT", "build.rs", "Cargo.toml
 indenter = "0.3.0"
 owo-colors = "4.3"
 autocfg = "1.0"
+anstream = "0.6.15"
```
pksunkara, line 23: [SUGGESTION] Let's update this to v1

color-eyre/examples/theme.rs
```diff
@@ -1,12 +1,13 @@
-use color_eyre::{Section, config::Theme, eyre::Report, owo_colors::style};
+use color_eyre::{Section, config::Theme, eyre::Report};
+use owo_colors::Style;
```
pksunkara, line 2: [SUGGESTION] Since we re-export `owo-colors`, we should probably recommend them to import it from `color_eyre`. So, keep the previous.

color-eyre/src/lib.rs
```diff
@@ -361,13 +361,16 @@ use std::sync::Arc;
...
+use anstream::AutoStream;
 use backtrace::Backtrace;
 pub use eyre;
 #[doc(hidden)]
 pub use eyre::Report;
 #[doc(hidden)]
 pub use eyre::Result;
 pub use owo_colors;
+use owo_colors::OwoColorize;
+use owo_colors::Style;
```
pksunkara, line 373: [NOTE] Combine these 2 lines into a single one.

color-eyre/src/lib.rs
```diff
@@ -458,3 +461,17 @@ pub enum ErrorKind<'a> {
...
     config::HookBuilder::default().install()
 }
+
+/// Apply owo_colors style if possible. Returns a string with/without ANSI
+/// escape symbols.
+fn style<S>(str: S, style: Style) -> String
+where
+    S: ToString + std::fmt::Display,
+{
+    let color_choice = AutoStream::choice(&std::io::stderr());
```
pksunkara, line 471: [ISSUE] Instead of deriving this every call, we should probably do this once at the beginning (maybe during `install`) and use it when called.

color-spantrace/src/lib.rs
```diff
@@ -374,3 +375,17 @@ impl fmt::Display for ColorSpanTrace<'_> {
...
     }
 }
+
+/// Apply owo_colors style if possible. Returns a string with/without ANSI
+/// escape symbols.
+fn style<S>(str: S, style: Style) -> String
+where
+    S: ToString + std::fmt::Display,
+{
+    let color_choice = AutoStream::choice(&std::io::stderr());
```
pksunkara, line 385: [ISSUE] 1. If we are deriving this here, why are we doing this again in `color_eyre`? Is it not better to unify both?
2. Is there no way to cache this?

This is a major blocker to finalize this PR. Since I will be pretty active, let's resolve this and merge it.

pksunkara (changes requested):
[ISSUE] Please add a small new test to color-eyre where we test with colors absent.

<sub>Reviewed with [tuicr](https://github.com/agavra/tuicr).</sub>

svenstaro:
@nikableh Could you take a look at the review?

PR #271, opened 2026-03-22
Rename the `eyre!` macro to `report!`

nik-rev:
Considering eyre plans to release 1.0 some time soon (https://github.com/eyre-rs/eyre/pull/269), it's a good time to clean up the API - to make the API leaner, help prevent choice overload, and make the macro names make more sense..

Having 2 macros that do the same thing leads to unnecessary churn in which name to use. That's why we should remove the `format_err!` macro.

On the other hand, `eyre!` macro is not a good name - a better name is `report!`, because a `Report` is created.

Renaming `ok_or_eyre` to `ok_or_report` (https://github.com/eyre-rs/eyre/pull/272) will mean we are consistent with the naming of `ok_or_report` method, and have more of a parallel between static error creation in the `None` case of an option, and **dynamic** error creation:

```rust
opt.ok_or_report("failed reading file")
```

```rust
opt.ok_or_else(|| report!("failed reading file: {path}"))
```


pksunkara:
I actually like the usage `Err(eyre!("cmd exited with non-zero status code"))` where as `report` in that doesn't seem intuitive

nik-rev:
> I actually like the usage `Err(eyre!("cmd exited with non-zero status code"))` where as `report` in that doesn't seem intuitive

Because this macro expands to the `Report` type, I'd say calling it `report!` makes the most sense. `eyre!` doesn't have any meaning to it, it's the name of the crate

Manishearth:
I don't think we should rename the macro. That's a pretty drastic change, and it's for a minor consistency benefit. eyre is a pretty popular library and I want migration to be easy.

If this were a "big" 1.0 I think I'd be in favor, but this isn't: we're cleaning up a few rough edges and releasing 1.0.

LeoniePhiline:
The changes we’ve lined up are relatively big, which shows that various contributors are very glad to finally have a breaking release which allows to _do things right_ and clean up properly, for a stable 1.0 which isn’t quickly followed by a 2.0. 

Thus, I would like to advocate for renaming the macro for the sake of discoverability:

It is much more obvious that an ad-hoc `Report` is created by a macro called by the same name, than by a macro called something entirely different.

This change is in line with the other changes which are already merged or prepared for merge.

These changes are consistent, they make sense, they make `eyre` mature, and they make the library easier for new users.

Migration is straightforward, changing `eyre!` to `report!`, including the corresponding imports.

If really necessary, a feature flag `backwards-compat` or `legacy-names` could allow users to keep using the old names until they find time to run a few regex search and replace commands across their code base. (But then, they could just stick with 0.x until they have time to upgrade.)

Manishearth:
I don't think the current changes are big. 99% of eyre users will not be affected by the breaking changes in 1.0. Renaming the macro _will_ affect everyone.

Having the breaking change be a no-op for most people is still valuable. It's the difference between being able to tweak an import (or let dependabot do it) and move on vs having to actually spend time looking through the changes.

For widely used crates, ease of migration does factor in in how quickly the ecosystem moves on.

I would be _very strongly_ opposed to this breaking change, with or without backcompat features. I would be weakly opposed to doing this compatibly (with `#[deprecated] pub use report as eyre`).

The reason behind the weak opposition is that I'm not really convinced of the change in the first place. The type is called `eyre::Report`, it makes sense that it is produced either by `eyre!()` or `report!()`. This library is well known enough that seeing `eyre!()` in a codebase won't be surprising.

So I'm not really convinced. I don't see a major difference, I see slight ones, and that's not enough for me to want to cause churn.

PR #272, opened 2026-03-22
Rename `ok_or_eyre` to `ok_or_report` in `OptionExt` trait

nik-rev:
Considering eyre plans to release 1.0 some time soon (https://github.com/eyre-rs/eyre/pull/269), it's a good time to clean up the API

After renaming `eyre!` to `report!` (https://github.com/eyre-rs/eyre/pull/271), renaming `ok_or_eyre` to `ok_or_report` will mean we are consistent with the naming of `ok_or_report` method, and have more of a parallel between static error creation in the `None` case of an option, and **dynamic** error creation:

```rust
opt.ok_or_report("failed reading file")
```

```rust
opt.ok_or_else(|| report!("failed reading file: {path}"))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,13 +231,13 @@ implies that you're creating a new error that saves the old error as its
 `source`. With `Option` there is no source error to wrap, so `wrap_err` ends up
 being somewhat meaningless.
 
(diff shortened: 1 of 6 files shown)


Manishearth:
I think we should do this if and only if we do https://github.com/eyre-rs/eyre/pull/271 as well.

PR #279, opened 2026-03-29
Bump edition to 2024

deckstose:
See https://github.com/eyre-rs/eyre/pull/276#issuecomment-4150983402.

Not sure if we should add safety comments to the examples, which now require an `unsafe` block for setting environment variables.


Cargo.toml
```diff
@@ -5,10 +5,11 @@ members = [
   "eyre",
   "simple-eyre",
 ]
+resolver = "3"
```
pksunkara: Isn't this the default in 2024? Do we need to specify it?

pksunkara (commented):
Why were `expr_2021` used in macros? Why can't they simply be `expr`?

pksunkara (approved):

Cargo.toml
deckstose: Apparently it doesn't work, I'm getting

```
warning: virtual workspace defaulting to `resolver = "1"` despite one or more workspace members being on edition 2024 which implies `resolver = "3"`
```

on nightly.

LeoniePhiline:
> Not sure if we should add safety comments to the examples, which now require an unsafe block for setting environment variables.

I think safety comments should be added. Examples are often used by less experienced developers and copy-pasted into their own projects.

A regular safety comment would present good, idiomatic style.

An added warning comment would hopefully make beginners aware of the `RUST_BACKTRACE` variable only being set for the sake of demonstration, with a recommendation to remove the `unsafe` block immediately when copying code.

That said, it might be better to remove the `set_var` altogether and instruct users to set it on the command line when running the backtrace examples.

PR #281, opened 2026-04-05
Move all dependencies to workspace and update some of them

oscargus:
Easier to keep track of versions if all are workspace dependencies. As seen from the first commit, this does not change `Cargo.lock`.

Second commit update some of the dependencies. Not sure if all support 1.85 though, so let's see what the tests say. `trybuild` is OK with 1.85 though.

The number of transitive dependencies is increased, primarily because the Unicode dependencies are split into multiple different dependencies. Also, trybuild now depends on toml. On the other hand, many old versions are removed, so the tests etc now relying on more up-to-date versions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


pksunkara:
This is not recommended way of doing things in not only cargo but also other package managers like node.js.

We only move a dependency to workspace if it's common in more than one crate.

PR #286, opened 2026-06-26
Eliminate pointer->reference->pointer during downcast

### LINKED ISSUE #285: Unsoundness in `Report::downcast_mut`
```rust
use eyre::Report;
use std::fmt;

#[derive(Debug)]
struct ErrorContext(&'static str);

impl fmt::Display for ErrorContext {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        fmt::Display::fmt(&self.0, f)
    }
}

fn main() {
    let mut error = Report::msg("inner error").wrap_err(ErrorContext("old context"));
    let context: &mut ErrorContext = error.downcast_mut().unwrap();
    context.0 = "new context";
    println!("{:?}", error);
}
```

Produces a miri complaint:

```
error: Undefined Behavior: trying to retag from <2375> for Unique permission at alloc920[0x18], but that tag only grants SharedReadOnly permission for this location
   --> /home/dfranke/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/non_null.rs:478:18
    |
478 |         unsafe { &mut *self.as_ptr() }
    |                  ^^^^^^^^^^^^^^^^^^^ this error occurs as part of retag at alloc920[0x18..0x28]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <2375> was created by a SharedReadOnly retag at offsets [0x18..0x28]
   --> eyre/src/error.rs:747:19
    = note: stack backtrace:
            0: std::ptr::NonNull::<ErrorContext>::as_mut::<'_>
                at /home/dfranke/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/non_null.rs:478:18: 478:37
            1: eyre::error::<impl eyre::Report>::downcast_mut::<ErrorContext>
                at eyre/src/error.rs:448:18: 448:43
            2: main
                at eyre/examples/downcast_mut.rs:15:38: 15:58
```

This is likely the same bug as https://github.com/dtolnay/anyhow/issues/451 (fixed yesterday) and https://github.com/zkat/miette/issues/469 (still open as of 2026-06-26).

dfoxfranke:
Based on <https://github.com/dtolnay/anyhow/pull/452/commits>. Fixes the same unsoundness in the same way. The included unit test fails under MIRI without the fix and passes afterward.

Fixes #285.

### DIFF
--- a/eyre/src/error.rs
+++ b/eyre/src/error.rs
@@ -605,7 +605,8 @@ where
 
 /// # Safety
 ///
-/// Requires layout of *e to match ErrorImpl<E>.
+/// Requires layout of *e to match ErrorImpl<E>. Ensures that returned pointee's layout
+/// matches `target`.
 unsafe fn object_downcast<E>(e: RefPtr<'_, ErrorImpl<()>>, target: TypeId) -> Option<NonNull<()>>
 where
     E: 'static,
@@ -622,7 +623,8 @@ where
 
 /// # Safety
 ///
-/// Requires layout of *e to match ErrorImpl<E>.
+/// Requires layout of *e to match ErrorImpl<E>. Ensures that returned pointee's layout
+/// matches `target`.
 unsafe fn object_downcast_mut<E>(
     e: MutPtr<'_, ErrorImpl<()>>,
     target: TypeId,
@@ -642,7 +644,8 @@ where
 
 /// # Safety
 ///
-/// Requires layout of *e to match ErrorImpl<ContextError<D, E>>.
+/// Requires layout of *e to match ErrorImpl<ContextError<D, E>>. Ensures that returned
+/// pointee's layout matches `target`.
 unsafe fn context_downcast<D, E>(
     e: RefPtr<'_, ErrorImpl<()>>,
     target: TypeId,
@@ -651,13 +654,16 @@ where
     D: 'static,
     E: 'static,
 {
+    let unerased = e.cast::<ErrorImpl<ContextError<D, E>>>();
     if TypeId::of::<D>() == target {
-        let unerased = e.cast::<ErrorImpl<ContextError<D, E>>>();
+        // Safety: Layout of `*unerased` matches `ErrorImpl<ContextError<D, E>>`.
         let addr = unsafe { ptr::addr_of!((*unerased.ptr.as_ptr())._object.msg) };
+        // Safety: `addr` is not null, layout matches `D`.
         Some(unsafe { NonNull::new_unchecked(addr.cast_mut()).cast::<()>() })
     } else if TypeId::of::<E>() == target {
-        let unerased = e.cast::<ErrorImpl<ContextError<D, E>>>();
+        // Safety: Layout of `*unerased` matches `ErrorImpl<ContextError<D, E>>`.
         let addr = unsafe { ptr::addr_of!((*unerased.ptr.as_ptr())._object.error) };
+        // Safety: `addr` is not null, layout matches `E`.
         Some(unsafe { NonNull::new_unchecked(addr.cast_mut()).cast::<()>() })
     } else {
         None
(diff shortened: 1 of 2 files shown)


eyre/src/error.rs
```diff
@@ -676,13 +676,13 @@ where
     E: 'static,
 {
     if TypeId::of::<D>() == target {
-        let unerased = unsafe { e.cast::<ErrorImpl<ContextError<D, E>>>().into_mut() };
-        let addr = NonNull::from(&unerased._object.msg).cast::<()>();
-        Some(addr)
+        let unerased = e.cast::<ErrorImpl<ContextError<D, E>>>();
+        let addr = unsafe { ptr::addr_of_mut!((*unerased.ptr.as_ptr())._object.msg) };
+        Some(unsafe { NonNull::new_unchecked(addr).cast::<()>() })
```
pksunkara, line 690: Should this not be `cast_mut`? (In all your changes too?)

pksunkara (commented):
CI is failing too.

eyre/src/error.rs
dfoxfranke: No, because in the one you called out it's already `*mut`. `cast_mut` changes a `*const T` to a `*mut T` without changing `T`; changing a `*mut T` to a `*mut U` is still spelled `cast` [on the primitive](https://doc.rust-lang.org/stable/std/primitive.pointer.html#method.cast-1), and also on [your wrapper](https://github.com/eyre-rs/eyre/blob/d33b9aa74b2338e6ed8deb51b7129792748dd998/eyre/src/ptr.rs#L133). Some of the other instances that are changed deal with `*const` pointers and have a `.cast_mut()`  to construct the `NonNull` that the return type requires. This is fine because stacked/tree-borrow semantics don't distinguish among raw pointers; the distinction between `*const` and `*mut` only exists in the type system. The important thing is keeping shared references out of the borrow chain.

dfoxfranke:
The CI failure doesn't look related to my patch. Is CI passing on master? Also, everything seems to be passing with both stable and nightly rustc but failing on beta, which is pretty weird.

pksunkara:
Looks like the CI failures were something that were caused by changes in beta. They were failing for nightly before and I marked them to be ignored to unblock things (https://github.com/eyre-rs/eyre/commit/4870bf9d49b0fc912dff2f0451f0344e84129ea7). Now, it moved to beta and I think it's time to really fix them.

dfoxfranke:
I tracked down the cause of the CI failures: https://github.com/eyre-rs/eyre/issues/287

dfoxfranke:
Marking as draft because I messed up the rebasing just now.

dfoxfranke:
Fixed.

Manishearth (commented):
Could you add safety comments on this?

dfoxfranke:
@Manishearth I'm having a go at addressing this request comprehensively and making sure that all unsafe code throughout the crate has its invariants properly documented. For the specific lines that my patch touches, the invariants are quite obvious. Elsewhere, they aren't, especially as pertains to the vtable protocol.  I don't enjoy encountering the word "sensible" in such specifications: https://github.com/eyre-rs/eyre/blob/20839201a1199df3d64c09ffa242594b08c6ac57/eyre/src/error.rs#L189-L200

I'd like to do a bit of refactoring to clarify this and tighten up the enforcement boundaries, but I'd rather keep this particular PR scoped just to addressing the specific unsoundness bug. I either can add a small commit here which narrowly satisfies your request for safety comments on the modified lines, or you could just merge this as-is and I'll sweep those comments into the follow-up work. Let me know your preference.

Manishearth:
I think narrowly addressing this would be fine. I didn't write this crate and any scaffolding to help evaluate new unsafe code would be great

dfoxfranke:
Okay, updated with safety comments.

Manishearth (approved):

PR #295, opened 2026-07-28
Fix: match `backtrace::capture` under v0 symbol mangling

phil-opp:
Follow-up to https://github.com/eyre-rs/eyre/pull/288

My hope is that this fixes the failing `test_error_backwards_compatibility` on macOS CI.


yaahc (approved):
lgtm, tyty!

